PR #116, opened 2019-06-27
Add JSON payload file support for send_transaction wctl command.

dowlandaiello:
Once again, congrats to the Perlin team for open-sourcing wavelet! I'm super excited for what the future holds.
I've been browsing the documentation to try to learn how to use Perlin both through the lens user interface and through wctl. However, I did have some trouble providing payloads for `transfer` transactions. It seems that, for now, wctl simply passes in a JSON file (`--payload` flag), but does not parse it (i.e. the entire file is sent as a payload, rather than each parameter). This pull request simply adds the ability to parse JSON payload files with the `send_transaction` wctl command.


tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+	"fmt"
+	"io/ioutil"
+
+	"github.com/perlin-network/wavelet/sys"
+	"github.com/pkg/errors"
+	"github.com/valyala/fastjson"
+)
+
+// Parser defines a generic JSON transaction payload parser.
+type Parser struct {
```
iwasaki-kenta: Possible to rename to `TransactionParserJSON`?

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+	Tag string // Transaction tag
+}
+
+var (
+	// ErrNoTag defines an error describing an empty TransactionTag.
```
iwasaki-kenta: Comments should be updated to `describing an empty Tag.`

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+	ErrNilField = errors.New("field is nil")
+
+	// ErrInvalidOperation defines an error describing an invalid operation value.
+	ErrInvalidOperation = errors.New("operation is invalid")
+)
+
+/* BEGIN EXPORTED METHODS */
+
+// NewParser initializes a new parser with the given transaction type.
+func NewParser(transactionTag string) *Parser {
```
iwasaki-kenta: `tag string`

dowlandaiello:
In addition to adding support for the `Transfer` tag, I've modularized the JSON payload parsing process by exporting all of the required methods to wavelet/tx_json.go.

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+
+/*
+	BEGIN TAG HANDLERS
+*/
+
+// parseTransfer parses a transaction payload with the transfer tag.
+func (parser *Parser) parseTransfer(data []byte) ([]byte, error) {
+	payload := bytes.NewBuffer(nil) // Initialize buffer
+
+	parsedData, err := parseJSONBytes(data) // Parse data
```
iwasaki-kenta: It might be better to in-line a lot of these methods such as `parseJSONBytes` and `parseRecipient` and `parseAmount` to avoid far too much encapsulation.

iwasaki-kenta: Would apply my comment above to all other functions below.

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+
+	transactions := parsedData.GetArray("payloads") // Get payloads
+
+	var batchLength [8]byte // Initialize batch length buffer
+
+	binary.LittleEndian.PutUint64(batchLength[:], uint64(len(transactions))) // Write batch length to buffer
+
+	payload.Write(batchLength[:4]) // Write batch length
+
+	for _, transaction := range transactions { // Iterate through transactions
```
iwasaki-kenta: `for _, tx := range transactions`

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+		json, err := transaction.StringBytes() // Get JSON
+		if err != nil {                        // Check for errors
+			return nil, err // Return found error
+		}
+
+		txParser := NewParser(string(transaction.GetStringBytes("tag"))) // Initialize parser
```
iwasaki-kenta: `parser := NewTransactionParserJSON(...)`

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+*/
+
+// parseOperation gets the operation of a particular stake transaction.
+func parseOperation(json *fastjson.Value) (byte, error) {
+	if !json.Exists("operation") { // Check no value
+		return byte(0), ErrNilField // Return nil field error
+	}
+
+	switch string(json.GetStringBytes("operation")) { // Handle different operations
+	case "0x00":
```
iwasaki-kenta: 0x00, 0x01, and 0x02 as strings can be simplified to a JSON number 0, 1, 2, 3, 4, ... etc. Would be good to validate that the operation code is less than `sys.TagBatch` as well, and that the number is not negative.

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+func parseFunctionPayload(json *fastjson.Value) ([8]byte, []byte, error) {
+	if !json.Exists("fn_payload") { // Check no value
+		return [8]byte{}, nil, ErrNilField // Return nil field error
+	}
+
+	var intBuf [8]byte // Initialize integer buffer
+
+	params := bytes.NewBuffer(nil) // Initialize payload buffer
+
+	for _, payloadValue := range json.GetArray("fn_payload") { // Iterate through payloads
```
iwasaki-kenta: I would recommend representing the payload in JSON more idiomatically like so:

```json
{
    "fn_payload": [
        {
            "type": "uint64",
            "value": 32
        },
        {
             "type": "hex",
             "value": "0adf39",
         }
    ]
}
```

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+	v, err := p.Parse(string(data)) // Parse json
+	if err != nil {                 // Check for errors
+		return nil, err // Return found error
+	}
+
+	return v, nil // Return parsed
+}
+
+// getValidTags gets a populated map of valid tags.
+func getValidTags() map[string]bool {
```
iwasaki-kenta: I would use `stringer` https://godoc.org/golang.org/x/tools/cmd/stringer and generate the string names for each individual operation type. The constant array of strings could go in the `sys` pkg under `const.go`, or otherwise in a separate file called `op.go`.

tx_json.go
dowlandaiello: Yup. Made more sense when it was just in its own package in cmd/wctl. Forgot to change it when it was moved to the root.

dowlandaiello: But yeah, totally agree.

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+
+// NewParser initializes a new parser with the given transaction type.
+func NewParser(transactionTag string) *Parser {
+	return &Parser{
+		Tag: transactionTag, // Set transaction tag
+	} // Initialize parser
+}
+
+// ParseJSON parses the given JSON payload input.
+func (parser *Parser) ParseJSON(data []byte) ([]byte, error) {
```
iwasaki-kenta: It would be good to validate the JSON using `fastjson`.

tx_json.go
```diff
@@ -0,0 +1,457 @@
+package wavelet
```
iwasaki-kenta: Rather than exposing a `Parser` insance, I believe what would be better is exporting a global function `ParseTransactionJSON(data []byte) (tag byte, payload []byte)`.

tx_json.go
```diff
@@ -0,0 +1,457 @@
...
+
+import (
```
iwasaki-kenta: Would recommend `quick.Check` tests to ensure that the JSON payloads for each individual tag/payload pair does not cause any panics. https://golang.org/pkg/testing/quick/

tx_json.go
dowlandaiello: I understand why parseJSONBytes might be inlined, but here it does feel like encapsulating parser methods for each different field provides some marginal benefit. Seems like inlining them would make things pretty messy (and a little hard to refactor later).

dowlandaiello: Not to mention you'd be dealing with a lot of unnecessary duplicate code.

sys/const.go
```diff
@@ -264,3 +265,12 @@ var (
...
 	}
 )
+
+// String converts a given tag to a string.
+func (tag Tag) String() string {
+	if tag < 0 || tag > 4 { // Check out of bounds
+		return "" // Return invalid tag
+	}
+
+	return []string{"TagNop", "TagTransfer", "TagContract", "TagStake", "TagBatch"}[tag] // Return tag
```
iwasaki-kenta: We should rename the string tag literals here to "nop", "transfer", "contract", "stake", "batch".

cmd/wctl/main.go
```diff
@@ -24,18 +24,20 @@ import (
...
+
+	"github.com/perlin-network/wavelet"
+	"github.com/perlin-network/wavelet/sys"
+	"github.com/perlin-network/wavelet/wctl"
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
 )
 
-var tagConversion = map[string]byte{
+var tagConversion = map[string]sys.Tag{
```
iwasaki-kenta: It would be better to either remove this map, or move it towards the `sys` pkg and keep the string naming consistent.

tx_json_test.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// TestParseJSON tests the functionality of the ParseJSON helper method.
+func TestParseJSON(t *testing.T) {
+	f := func(jsonData []byte, tag string) bool {
+		payload, err := ParseJSON(jsonData, tag) // Attempt to parse
+		return !(err != nil && payload != nil)   // Check errored but still returned
+	}
+
+	if err := quick.Check(f, nil); err != nil { // Check for errors
```
iwasaki-kenta: `assert.NoError(quick.Check(f, nil))`.

tx_json.go
```diff
@@ -0,0 +1,415 @@
...
+		return nil, ErrNoTag // Return no tag error
+	}
+
+	if err := fastjson.ValidateBytes(data); err != nil { // Check not valid JSON
+		return nil, err // Return found error
+	}
+
+	valid := false // Initialize valid buffer
+
+	for i := 0; i < 5; i++ { // Iterate through tags
```
iwasaki-kenta: We could check if the tag is really a tag using the `sys` pkg.

tx_json.go
```diff
@@ -0,0 +1,415 @@
...
+
+			tag = sys.Tag(i).String() // Set tag to formal name
+		}
+	}
+
+	if !valid { // Check invalid
+		return nil, ErrInvalidTag // Return error
+	}
+
+	switch tag { // Handle different tag types
```
iwasaki-kenta: I believe we should make `TagToLabels` and `LabelsToTags` map in the `sys` pkg so we don't have to repeat the actual label values here.

tx_json.go
```diff
@@ -0,0 +1,415 @@
...
+
+	var amount [8]byte                                      // Initialize integer buffer
+	binary.LittleEndian.PutUint64(amount[:], decodedAmount) // Write to buffer
+
+	_, err = payload.Write(amount[:]) // Write amount value
+	if err != nil {                   // Check for errors
+		return nil, err // Return found error
+	}
+
+	if json.Exists("gas_limit") { // Check has gas limit value
```
iwasaki-kenta: Could you declare all of these string constants at the top of the file?

tx_json.go
dowlandaiello: Seems like the `sys.TagConversiono` map already addresses this issue. Could you elaborate a bit more on what you mean?

iwasaki-kenta: Ah yep, in that case it's a matter of not having to repeat the label values "nop", "transfer", etc. by using the `sys.TagConversion` map or `String()` method for tags.

PR #121, opened 2019-06-29
Changed to urfave/cli for the shell, added autocompletion, some 80 col changes.

diamondburned:
- Changed shell command evaluation to `urfave/cli`
    - Command help is different
    - Quotes are done properly now
- Added autocompletion
- Changed `shell.go` to `cli.go`
- Moved the actions to `actions.go` so `cli.go` isn't too long
- `cmd/wavelet` files have some lines changed so they're not too long
- Fixed `cmd/wavelet/README.md`
- Tidied `go.mod`


cmd/wavelet/cli.go
```diff
@@ -0,0 +1,239 @@
...
+	client *skademlia.Client
+	ledger *wavelet.Ledger
+	logger zerolog.Logger
+	keys   *skademlia.Keypair
+
+	completion []string
+}
+
+func NewCLI(client *skademlia.Client, ledger *wavelet.Ledger, keys *skademlia.Keypair) (*CLI, error) {
+	CLI := &CLI{
```
iwasaki-kenta: Would be good to lower-case it to `cli`.

cmd/wavelet/completion.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		return
+	}
+
+	return readline.PcItemDynamic(fn)
+}
+*/
+
+func (cli *CLI) getCompleter() *readline.PrefixCompleter {
```
iwasaki-kenta: Think this function could be inlined,

iwasaki-kenta: Will cleanup after merging in the autocomplete PR though.

cmd/wavelet/cli.go
diamondburned: I thought about this too, but the `urfave/cli` library uses `cli` as its package name, so it'll have to be something else.

diamondburned: Nevermind, fixed

cmd/wavelet/completion.go
diamondburned: If this is to be inlined, it'll look like this 
```go
	for _, cmd := range c.app.Commands {
		completers = append(completers, readline.PcItem(
			cmd.Name, readline.PcItemDynamic(func(string) []string {
				return c.completion
			}),
		))

		for _, alias := range cmd.Aliases {
			completers = append(completers, readline.PcItem(
				alias, readline.PcItemDynamic(func(string) []string {
					return c.completion
				}),
			))
		}
	}
``` 

in `cli.go`

iwasaki-kenta: Gotcha; no problem though since the completers will be manually computed from `ledger.Find()`.

diamondburned: Alright, I pushed a commit to inline it anyway, since it could be cleaner than the code above.

PR #125, opened 2019-07-09
Ledger state syncing / pull missing transactions - minor cleanup and improvements.

a-urth:
Stop pulling missing transactions in case state sync needs to be done immediately instead of waiting for now more missing transaction
Lower most of syncing related timeouts - pulling, querying etc
Add check for live connections in most places - before pulling and for `SelectPeers`
Change how gossip network part works - do not store streams, create new stream each time instead and just pass in case of error - it avoid problem using "dead" connection and connection which were not removed from map. And since we send transactions in batch overhead of creating new connection is not that big.

### DIFF
--- a/gossip.go
+++ b/gossip.go
@@ -24,8 +24,6 @@ import (
 	"github.com/perlin-network/noise/skademlia"
 	"github.com/perlin-network/wavelet/debounce"
 	"github.com/perlin-network/wavelet/log"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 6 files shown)


a-urth:
@losfair cleaned up code from comments and prints, added description to what was actually done within this PR.
Syncing/snowball changes as discussed will be done in separate PR
@iwasaki-kenta fyi

a-urth:
@losfair or @iwasaki-kenta if its ok can we please merge it? since there are some changes in here which i want to use in order to test syncing fix in other branch?

iwasaki-kenta:
> @losfair or @iwasaki-kenta if its ok can we please merge it? since there are some changes in here which i want to use in order to test syncing fix in other branch?

Looks good, LGTM.

PR #126, opened 2019-07-10
Fixed seed conditions and added tests for applying transactions.

losfair:


accounts.go
```diff
@@ -30,6 +30,10 @@ import (
 	"unsafe"
 )
 
+type LikeAccounts interface {
```
iwasaki-kenta: Would be better to expose a `avl.Snapshottable` interface in the `avl` pkg.

graph.go
```diff
@@ -718,3 +720,238 @@ func (g *Graph) validateTransactionParents(tx *Transaction) error {
...
+
+	for _, rw := range rws {
+		balance, _ := ReadAccountBalance(snapshot, rw.account)
+		WriteAccountBalance(snapshot, rw.account, balance+rw.amount)
+
+		snapshot.Delete(rw.Key())
+	}
+}
+
+func (g *Graph) RewardValidators(snapshot *avl.Tree, root *Transaction, tx *Transaction, logging bool) error {
```
iwasaki-kenta: Why move validator logic into `graph.go`? Also, would be good to hide/expose it s.t. `rewardValidators`.

graph.go
```diff
@@ -718,3 +720,238 @@ func (g *Graph) validateTransactionParents(tx *Transaction) error {
...
+			Hex("rewardee_tx_id", rewardee.ID[:]).
+			Hex("entropy", entropy).
+			Float64("acc", acc).
+			Float64("threshold", threshold).Msg("Rewarded validator.")
+	}
+
+	return nil
+}
+
+func (g *Graph) CollapseTransactions(accounts LikeAccounts, round uint64, latestRound *Round, root *Transaction, end *Transaction, logging bool) (*CollapseResults, error) {
```
iwasaki-kenta: Again, would be good to hide/not expose it s.t. `collapseTransactions`. Also, why move it to `graph.go`?

tx.go
```diff
@@ -91,12 +91,20 @@ func NewBatchTransaction(creator *skademlia.Keypair, tags []byte, payloads [][]b
...
 }
 
-func AttachSenderToTransaction(sender *skademlia.Keypair, tx Transaction, parents ...*Transaction) Transaction {
+func AttachSenderToTransaction(sender *skademlia.Keypair, _tx *Transaction, parents ...*Transaction) *Transaction {
+	tx := *_tx
+	AttachSenderToTransactionInPlace(sender, &tx, parents...)
+	return &tx
+}
+
+func AttachSenderToTransactionInPlace(sender *skademlia.Keypair, tx *Transaction, parents ...*Transaction) {
```
iwasaki-kenta: Rename to `AppendSenderToTransaction` to imply that it in-place modifies the transaction provided. Also please provide documentation on these functions.

graph.go
iwasaki-kenta: To clarify, anything to do with rewarding/collapsing/staking should not belong in `graph.go`. Anything that modifies ledger state should remain in `ledger.go`, unless you have good reasons (say for testing infrastructure for example) to move it to `graph.go`.

graph.go
```diff
@@ -612,7 +614,7 @@ func (g *Graph) deleteProgeny(id TransactionID) {
 	}
 }
 
-func (g *Graph) validateTransaction(tx Transaction) error {
+func (g *Graph) validateTransaction(tx *Transaction) error {
```
iwasaki-kenta: What about validating incoming transactions parent seeds? We need to ensure that parent seeds are properly validated.

tx.go
```diff
@@ -114,26 +122,44 @@ func AttachSenderToTransaction(sender *skademlia.Keypair, tx Transaction, parent
...
+		hasher, err := blake2b.New(32, nil)
+		if err != nil {
+			panic(err) // should never happen
+		}
 
-	t.Seed = blake2b.Sum256(buf)
-	t.SeedLen = byte(prefixLen(t.Seed[:]))
+		_, err = hasher.Write(t.Sender[:])
+		if err != nil {
+			panic(err) // should never happen
```
iwasaki-kenta: If it should never happen, then write instead _, _ = hasher.Write(t.Sender[:]).

iwasaki-kenta: Same goes for other error-check statements within `rehash()`.

tx.go
```diff
@@ -228,6 +259,15 @@ func UnmarshalTransaction(r io.Reader) (t Transaction, err error) {
 		}
 	}
 
+	t.ParentSeeds = make([]TransactionSeed, len(t.ParentIDs))
```
iwasaki-kenta: Please update the web documentation to take into account the inclusion of parent seeds within each and every transaction.

tx_applier.go
```diff
@@ -34,6 +34,40 @@ type ContractExecutorState struct {
 	GasLimit uint64
 }
 
+func ApplyTransaction(round *Round, snapshot *avl.Tree, tx *Transaction) error {
+	original := snapshot.Snapshot()
```
iwasaki-kenta: Rename `snapshot` to `state` s.t. `state.Snapshot()`.

tx_applier_test.go
```diff
@@ -0,0 +1,191 @@
...
+			fromAccount.effect.Balance -= amount
+			toAccount.effect.Balance += amount
+
+			payload := bytes.NewBuffer(nil)
+			payload.Write(toAccountID[:])
+			var intBuf [8]byte
+			binary.LittleEndian.PutUint64(intBuf[:], amount)
+			payload.Write(intBuf[:])
+
+			tx := AttachSenderToTransaction(fromAccount.keys, NewTransaction(fromAccount.keys, sys.TagTransfer, payload.Bytes()))
```
iwasaki-kenta: I do not see any tests for other transaction types apart from `sys.TagTransfer` transactions.

tx.go
a-urth: i would put `log` even in case it will "never" happen.

accounts.go
a-urth: in go usually you name interfaces in the form 'er, like Snapshotter :D

iwasaki-kenta: True - actually this interface in particular doesn't appear to be needed from what I see. The point of this interface is to allow for mocking in `tx_applier_test.go`, when what could be done instead is just struct embedding over `avl.Tree` and overriding the Snapshot() method.

tx.go
losfair: If something that should never happen does happen, it should be a fatal error. Panic in this case.

tx_applier_test.go
losfair: Also sys.TagStake. These tests are only for general transaction applying routines, but not for specific tags.

graph.go
losfair: For testing, we should not need to initialize a full ledger instance just to test transaction processing.

graph.go
losfair: Same as above.

tx.go
iwasaki-kenta: I'd argue just explicitly ignoring via _, _ = hasher.Write(t.Sender[:]). The documentation states that an error will never be thrown. A panic nor an error in this case should never have to occur. If needed, a comment can be appended to emphasize this.

graph.go
iwasaki-kenta: This validation would need to be done in `validateTransactionParents` in this case.

iwasaki-kenta: A test must be adjusted to check for this in `graph_test.go`.

tx.go
losfair: I don't think Append and Attach really imply this difference.

iwasaki-kenta: The `Append` prefix pattern, I am taking it as a style guideline from the `fasthttp`, `fastjson` etc libraries that provide functions which modify some pointer in-place.

tx_applier_test.go
iwasaki-kenta: We should test not only for general transaction applying routines. The most important and sensitive logic we should whitebox test for example is gas fee deduction, and recursively calling a smart contract function.

graph.go
iwasaki-kenta: `graph.go` should be purely in-memory however and be entirely algorithmic. The functions moved to `graph.go` in this case modify ledger state.

`graph.go` already handles in-memory indices, managing the directed acyclic graph structure, and validating transactions. It's best not to have it handle actually applying transactions to a ledger state too.

tx_applier_test.go
losfair: These will need calling a full smart contract - better do this in another PR.

graph.go
```diff
@@ -708,6 +704,10 @@ func (g *Graph) validateTransactionParents(tx *Transaction) error {
 		if maxDepth < parent.Depth { // Update max depth witnessed from parents.
 			maxDepth = parent.Depth
 		}
+
+		if parent.Seed != tx.ParentSeeds[i] {
```
iwasaki-kenta: Could you add a test for checking that validateTransactionParents validates parent seeds as expected?

tx.go
```diff
@@ -91,12 +91,20 @@ func NewBatchTransaction(creator *skademlia.Keypair, tags []byte, payloads [][]b
 	return NewTransaction(creator, sys.TagBatch, append([]byte{byte(len(tags))}, buf...))
 }
 
-func AttachSenderToTransaction(sender *skademlia.Keypair, tx Transaction, parents ...*Transaction) Transaction {
+func AttachSenderToTransaction(sender *skademlia.Keypair, _tx *Transaction, parents ...*Transaction) *Transaction {
```
a-urth: why `tx` is prefixed with underscored? i don't think in go there is such naming convention.

tx.go
```diff
@@ -91,12 +91,20 @@ func NewBatchTransaction(creator *skademlia.Keypair, tags []byte, payloads [][]b
...
+	tx := *_tx
```
a-urth: why creating new transaction, sending it by pointer and returning new pointer? since this function already takes pointer its kinda obvious that it will be mutated.
maybe we should follow on of the two:
- either just mutate object coming from parameters and do not return it
- return value from function (sender and parents in this context) and attach them to transaction on upper level

current version is much more confusing than with no new structure returned
what are we caring about - less allocation/speed or immutability/safety(arguably)? lets decide one style and follow it.

iwasaki-kenta:
FIXME(@iwasaki-kenta): Whitepaper must be updated with respect to the new seed conditions.

a-urth:
One more thing on topic about pointers vs values. It would've been good to have benchmark if we care about speed. Because every struct value which is not passed by pointer is allocated on the stack which is much faster then the heap. So maybe its actually cheaper to allocate new struct for each function call on the stack then to allocate it on the heap and then resolve pointer (which is also not for free).

iwasaki-kenta:
> One more thing on topic about pointers vs values. It would've been good to have benchmark if we care about speed. Because every struct value which is not passed by pointer is allocated on the stack which is much faster then the heap. So maybe its actually cheaper to allocate new struct for each function call on the stack then to allocate it on the heap and then resolve pointer (which is also not for free).

I completely agree as well. Originally transactions were only heap-allocated should they be added to the graph of a node.

We should setup performance regression tests in the near future. I will also adjust some styling in this PR after merging it in to master given that all tests pass and nothing breaks.

tx.go
iwasaki-kenta: Fixing it in master.

tx.go
iwasaki-kenta: I agree as well. I would prefer  we care more about immutability/safety to be sure that the code itself is self-describing, much like idiomatic Go in Go's standard library.

PR #129, opened 2019-07-16
cmd/wavelet: fix unspecified wallet flag not generating random wallet

hasyimibhar:
The documentation states that by not specifying the `--wallet` flag, a new wallet will be [randomly generated](https://wavelet.perlin.net/docs/setup#wallet-management). But currently, not specifying the `--wallet` flag will make Wavelet use `config/wavelet.txt` as the wallet. In order to to generate a random wallet, the `--wallet` flag has to be explicitly set to empty, e.g.:

```sh
$ ./wavelet --port 3000 --api.port 9000 --wallet ""
```

This PR fixes this bug, so that not specifying the `--wallet` flag will generate a random wallet. It also adds integration tests for `cmd/wavelet`. It requires `wavelet` binary to be in `$PATH`, so you need to install it if you want to run the test locally:

```sh
$ go install ./cmd/wavelet
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
         - "1.12.5"
       install: true
       script:
+        - GO111MODULE=on go build -o $GOPATH/bin/wavelet ./cmd/wavelet
         - GO111MODULE=on go test -coverprofile=coverage.txt -covermode=atomic -bench -race ./...
       after_success:
         - bash <(curl -s https://codecov.io/bash)
--- a/cmd/wavelet/main.go
+++ b/cmd/wavelet/main.go
@@ -98,7 +98,6 @@ func main() {
 		}),
 		altsrc.NewStringFlag(cli.StringFlag{
 			Name:   "wallet",
-			Value:  "config/wallet.txt",
 			Usage:  "Path to file containing hex-encoded private key. If the path specified is invalid, or no file exists at the specified path, a random wallet will be generated. Optionally, a 128-length hex-encoded private key to a wallet may also be specified.",
 			EnvVar: "WAVELET_WALLET",
 		}),
--- a/cmd/wavelet/main_test.go
+++ b/cmd/wavelet/main_test.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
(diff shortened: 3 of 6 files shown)


hasyimibhar:
`TestLimiterTimer` seems to have intermittent failures. Even on my local machine it sometimes fail.

a-urth:
@hasyimibhar yes, kinda, it fails rarely though. Its non deterministic since it relies on timeout and on unknown factor which is speed of the system you running tests on.
i've noticed it couple of time too (literally 2 times so far), but anyway, thas my bad. i'll increase timeout to be 100% sure that it not conflicts with timeout.

iwasaki-kenta:
Rather than assuming the Wavelet binary is in `PATH`, is it possible instead to just directly test the functions that make up the CLI?

I reckon that would be better, and make CI setup easier.

hasyimibhar:
Yeah it should be possible, and I think it's better as you can capture the test coverage as well. One way is to move all of the cli setup code outside of `main` and into its own function, so that what's left in `main` is something like this:

```go
func main() {
    cmd := &Command{}
    os.Exit(cmd.Run(os.Args))
}
```

Then you can test `cmd/wavelet` like this:

```go
func TestWavelet_Default(t *testing.T) {
        port := nextPort(t)
        apiPort := nextPort(t)

	cmd := &Command{}
	go runWavelet(t, cmd, []string{"--port", port, "--api.port", apiPort})
        defer cmd.Shutdown()

        // Wait for Wavelet to finish setup
        <-cmd.Wait()

        // Helper test method to call GET /ledger
	ledger := getLedgerStatus(t, apiPort)

	// Make sure a random wallet is generated
	assert.NotEqual(t, wallet1, ledger.PublicKey)
	assert.NotEqual(t, wallet2, ledger.PublicKey)
	assert.NotEqual(t, wallet3, ledger.PublicKey)
}

func runWavelet(t *testing.T, cmd *Command, args []string) {
    exitCode := cmd.Run(args)
    if exitCode != 0 {
        t.Fatal("non-zero exit code")
    }
}
```

But I noticed there's a major change in `cmd/wavelet` pending in #121, so I think it's better to wait for that PR to be merged first before proceeding.

hasyimibhar:
This PR is obsolete. Refer to #146 instead.

PR #130, opened 2019-07-17
Fix debounce limiter timer related test.

a-urth:


a-urth:
@hasyimibhar good job on trying this many times! my bad on rushing things and not testing it properly and not fixing it properly as a result. there was race condition (in test, how `called` was calculated), which i was aware about in the beginning but to lazy to fix and forgot now.
Now it actually works.

hasyimibhar:
I tested it again and it doesn't fail anymore. Looks good. :+1:

PR #132, opened 2019-07-18
Prefer values over heap-allocated values, simplify memory model, and fix appending parent transaction seeds to transactions.

iwasaki-kenta:
api, cmd/wavelet, gossip, collapse, genesis, ledger, round: revert to original memory alloc model such that values are emphasized over pointers to ensure correctness
tx: fix AttachSenderToTransaction to properly sort and index parent seeds with respect to index locations of parent ids
tests: parallelize tests where possible

### DIFF
--- a/collapse.go
+++ b/collapse.go
@@ -23,6 +23,13 @@ func processRewardWithdrawals(round uint64, snapshot *avl.Tree) {
 	}
 }
 
+// rewardValidators deducts a transaction fee from a transactions creator, and transfers the fee
+// to a rewardee which is determined by a validator reward scheme given the selected ancestry
+// of the transaction.
+//
+// If no rewardee is selected, then the transaction fee is simply burned. A reference to
+// a transaction is expected when calling this function to prevent any additional requirements
+// of looking up said transaction within the graph.
 func rewardValidators(g *Graph, snapshot *avl.Tree, tx *Transaction, logging bool) error {
 	fee := sys.TransactionFeeAmount
 
--- a/collapse.go
+++ b/collapse.go
@@ -23,7 +23,7 @@ func processRewardWithdrawals(round uint64, snapshot *avl.Tree) {
 	}
 }
 
-func rewardValidators(g *Graph, snapshot *avl.Tree, root *Transaction, tx *Transaction, logging bool) error {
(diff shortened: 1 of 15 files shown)


collapse.go
```diff
@@ -23,7 +23,7 @@ func processRewardWithdrawals(round uint64, snapshot *avl.Tree) {
 	}
 }
 
-func rewardValidators(g *Graph, snapshot *avl.Tree, root *Transaction, tx *Transaction, logging bool) error {
+func rewardValidators(g *Graph, snapshot *avl.Tree, tx *Transaction, logging bool) error {
```
a-urth: so are we passing by value or by pointer? i though that idea is to pass values everywhere when we do not need explicitly mutate state to prevent potential errors.
in this particular case, i guess we can pass only data used inside, instead of whole transaction

tx.go
```diff
@@ -117,101 +114,98 @@ func AppendSenderToTransaction(sender *skademlia.Keypair, tx *Transaction, paren
...
 	tx.Sender = sender.PublicKey()
 	tx.SenderSignature = edwards25519.Sign(sender.PrivateKey(), tx.Marshal())
 
 	tx.rehash()
+
+	return tx
 }
 
-func (t *Transaction) rehash() {
+func (tx *Transaction) rehash() *Transaction {
```
a-urth: why this still returns pointer? since its method of transaction struct itself i guess it can mutate. otherwise it needs to return value.

collapse.go
iwasaki-kenta: For this case we pass it by pointer because we can assume that all transactions called with rewardValidators() must be stored within the graph, and that all functions that call rewardValidators() already holds a reference to the pointer to the tx transaction.

tx.go
iwasaki-kenta: True, will remove that.

collapse.go
a-urth: if graph has all transaction then passing tx.ID should be sufficient right? because you can take it's content from the graph.

iwasaki-kenta: That is true, the point is to prevent a lookup to optimize performance since rewardValidators() by itself is already very expensive. Will document it.

PR #134, opened 2019-07-19
State and transaction syncing - fixes and flow change.

a-urth:
1. Make snowball generic to work with interface, fix tests and change existing code according to that
2. Change out of sync handler so it takes round index and decides whether given round is out of sync
3. Change how sync works - count votes with results in form of `true`/`false` which represents whether we out of sync according to peer's state
4. Change timeout for syncing
5. Lower timeouts for query and sync grpc requests
6. Fix depth of missing parent
7. ~Add buffer to store added transaction while node is out of sync, start node as out of sync one until opposite will be proved. Add all transactions from buffer upon sync~.
Add error and prevent node from creating transaction if node is not synced. Node is considered synced if it has synced state and has at least one successful finalization.
8. Brash up graph tests a bit
9. Add tests for `vote.go`

Still working on
1. ~In case transaction is deleted from `missing` while pruning, if it has children which are incomplete, they are never going to be completed.~ since update root happens on each sync/finalization and it resolves missing/incomplete transaction, this one right here should not be the case
2. ~In case transaction is deleted from `missing` while updating root, if it has children which are incomplete, they are never going to be completed.~
3. ~Seems like sometimes after restart node is not receiving gossiped transactions~ i'm pretty confident that its because noise/grpc, but will try to investigate more


graph.go
```diff
@@ -538,7 +561,13 @@ func (g *Graph) updateGraph(tx *Transaction) error {
 		g.metrics.receivedTX.Mark(int64(tx.LogicalUnits()))
 	}
 
-	for _, childID := range g.children[tx.ID] {
+	g.completeChildren(tx.ID)
+
+	return nil
+}
+
+func (g *Graph) completeChildren(parentID TransactionID) {
```
iwasaki-kenta: Care to rename to `resolveChildren`?

graph.go
```diff
@@ -294,12 +303,18 @@ func (g *Graph) PruneBelowDepth(targetDepth uint64) int {
 	}
 
 	for id, depth := range g.missing {
-		if depth > targetDepth {
-			continue
+		if depth <= targetDepth {
+			delete(g.missing, id)
+			count++
```
iwasaki-kenta: Care to put count++ after the two delete() statements?

graph.go
```diff
@@ -222,12 +225,21 @@ func (g *Graph) UpdateRootDepth(rootDepth uint64) {
 	g.rootDepth = rootDepth
 
 	for id, depth := range g.missing {
-		if rootDepth <= sys.MaxDepthDiff+depth {
-			continue
+		if rootDepth > sys.MaxDepthDiff+depth {
+			delete(g.missing, id)
+			g.completeChildren(id)
```
iwasaki-kenta: Will be good to have a line break after delete(g.missing, id).

graph.go
a-urth: not sure about children, since its just reference to transaction's children, not another transaction

a-urth:
Preventing from creating transactions on the node in case of desync is not reliable. Since its possible to do finalization after sync, but it takes same time which takes other nodes to proceed to next 2 rounds which puts node in desync state again and transaction is getting lost.
Its better than nothing, but still sometimes transaction can be created but won't get applied by the network.

a-urth:
@rkeene @iwasaki-kenta guys, there is crucial improvement and bug fix within this pr. Lets come to some resolution with it.
Since @hasyimibhar is now working on state syncing as well and we are getting closer to next release.

iwasaki-kenta:
> @rkeene @iwasaki-kenta guys, there is crucial improvement and bug fix within this pr. Lets come to some resolution with it.
> Since @hasyimibhar is now working on state syncing as well and we are getting closer to next release.

Still not 100% sure about making Snowball take in interfaces and having strings generated via fmt.Sprintf for a rounds identity is necessarily the best option.

In this case, let's put up an issue to address cleaning up/refactoring away interface{}'s, and proceed to merge with master as there are many valuable fixes in this PR.

PR #135, opened 2019-07-22
Validate WASM code for smart contracts before writing it into account state.

losfair:


contract.go
```diff
@@ -241,6 +251,10 @@ func (e *ContractExecutor) Execute(snapshot *avl.Tree, id AccountID, round *Roun
 		return errors.Wrapf(ErrContractFunctionNotFound, `fn "_contract_%s" does not exist`, name)
 	}
 
+	if vm.FunctionCode[entry].NumParams != 0 {
```
iwasaki-kenta: Why must the entry function not have parameters?

losfair: That's our contract ABI:

https://github.com/perlin-network/smart-contract-rs/blob/master/smart-contract-macros/src/lib.rs#L168
https://github.com/perlin-network/smart-contract-rs/blob/master/smart-contract-macros/src/lib.rs#L185

PR #137, opened 2019-07-23
utils: add test for SelectPeers

hasyimibhar:
Added `TestSelectPeers`. Also added test utility `newNode`, which might be useful for other tests.

### DIFF
--- a/utils_test.go
+++ b/utils_test.go
@@ -1,0 +1,112 @@
+package wavelet
+
+import (
+	"net"
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/perlin-network/noise"
+	"github.com/perlin-network/noise/cipher"
+	"github.com/perlin-network/noise/handshake"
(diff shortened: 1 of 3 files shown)


utils_test.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}()
+
+	return client, addr, func() {
+		server.GracefulStop()
+		cleanup()
+	}
+}
+
+func newKV(kv string, path string) (store.KV, func()) {
```
iwasaki-kenta: I think we should expose this as a function within the `store` pkg. Thoughts?

hasyimibhar: Yeah we should. I actually copy-pasted this from the `avl` package: https://github.com/perlin-network/wavelet/blob/master/avl/tree_test.go#L431-L454

iwasaki-kenta: Awesome, care to do that in this PR as well?

hasyimibhar: Alright sure.

PR #138, opened 2019-07-23
tx_parser: add tests

hasyimibhar:
Add tests for `Parse*Transaction` functions.

### DIFF
--- a/tx_parser_test.go
+++ b/tx_parser_test.go
@@ -1,0 +1,453 @@
+package wavelet
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"testing"
+
+	"github.com/perlin-network/noise/skademlia"
+	"github.com/perlin-network/wavelet/sys"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseTransferTransaction(t *testing.T) {
+	tx := validTransfer(t)
+	payload := encodeTransfer(tx)
+	tx2, err := ParseTransferTransaction(payload)
+	assert.NoError(t, err)
+	assert.Equal(t, tx, tx2)
+
+	// FuncParams is optional
+	txNoFuncParams, err := ParseTransferTransaction(payload[:SizeAccountID+8+8+8+4+len(tx.FuncName)])
+	assert.NoError(t, err)
+	tx.FuncParams = nil
+	assert.Equal(t, tx, txNoFuncParams)
+
+	// FuncName is optional
+	txNoFuncName, err := ParseTransferTransaction(payload[:SizeAccountID+8+8+8])
+	assert.NoError(t, err)
+	tx.FuncName = nil
+	assert.Equal(t, tx, txNoFuncName)
+
+	// GasDeposit is optional
+	txNoGasDeposit, err := ParseTransferTransaction(payload[:SizeAccountID+8+8])
+	assert.NoError(t, err)
+	tx.GasDeposit = 0
+	assert.Equal(t, tx, txNoGasDeposit)
+
+	// GasLimit is optional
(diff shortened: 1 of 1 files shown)


tx_parser_test.go
```diff
@@ -0,0 +1,442 @@
...
+		},
+		Payloads: [][]byte{
+			encodeTransfer(validTransfer(t)),
+			encodeStake(validStake(t, sys.PlaceStake)),
+			encodeContract(validContract(t)),
+		},
+	}
+}
+
+func encodeTransfer(tx Transfer) []byte {
```
hasyimibhar: I think these `encode*` functions can be converted into member function of the payload struct, e.g.:

```go
func (t Transfer) Marshal() []byte {
    // blah
}
```

This would allow `wavelet/cmd` to be simplified, e.g.:

```go

func (cli *CLI) pay(cmd []string) {
	if len(cmd) != 2 {
		fmt.Println("pay <recipient> <amount>")
		return
	}

	recipient, err := hex.DecodeString(cmd[0])
	if err != nil {
		cli.logger.Error().Err(err).Msg("The recipient you specified is invalid.")
		return
	}

	if len(recipient) != wavelet.SizeAccountID {
		cli.logger.Error().Int("length", len(recipient)).Msg("You have specified an invalid account ID to find.")
		return
	}

	amount, err := strconv.ParseUint(cmd[1], 10, 64)
	if err != nil {
		cli.logger.Error().Err(err).Msg("Failed to convert payment amount to a uint64.")
		return
	}

	snapshot := cli.ledger.Snapshot()

	var recipientID wavelet.AccountID
	copy(recipientID[:], recipient)

	balance, _ := wavelet.ReadAccountBalance(snapshot, cli.keys.PublicKey())
	_, codeAvailable := wavelet.ReadAccountContractCode(snapshot, recipientID)

	if balance < amount+sys.TransactionFeeAmount {
		cli.logger.Error().Uint64("your_balance", balance).Uint64("amount_to_send", amount).Msg("You do not have enough PERLs to send.")
		return
	}

        payload := wavelet.Transfer{
            Recipient: recipient,
            Amount: amount,
        }

	if codeAvailable {
                payload.GasLimit = balance-sys.TransactionFeeAmount
                payload.FuncName = []byte("on_money_received")
	}

	tx, err := cli.sendTransaction(wavelet.NewTransaction(cli.keys, sys.TagTransfer, payload.Marshal()))
	if err != nil {
		return
	}

	cli.logger.Info().Msgf("Success! Your payment transaction ID: %x", tx.ID)
}
```

Maybe we can also introduce helper methods for creating each transaction type, e.g.:

```go
// This will simply call wavelet.NewTransaction(keys, sys.TagTransfer, payload.Marshal())
tx, err := wavelet.NewTransferTransaction(keys, wavelet.Transfer{Recipient: recipient, Amount: 1234})
```

iwasaki-kenta: Completely agree on that - that sounds great :smile:.

hasyimibhar:
Rebased to master and updated tests.

tx_parser_test.go
hasyimibhar: It will also affect `ParseJSON` since it is also building the payload. I think it's better to do it in a separate PR since there will be a lot of changes.

a-urth:
Please, do not rebase from master, just merge. Since Kenta doesn't want to squash when merging PRs to master, it will be commit hell in master with rebasing to feature branches.

hasyimibhar:
@a-urth I see, I thought all PRs are squashed because I took a glance at the commit log in `master` and it looks like each PR are squashed.

a-urth:
looks like some were squashed, you're right, but most aren't, and thats our approach as of now, even i don't really like it :D

PR #139, opened 2019-07-24
tx_parser, tx_applier: refactor payload marshaling

hasyimibhar:
Refactors payload marshaling into `Marshal` function, as it's used several times within `wavelet` package.


hasyimibhar:
Seems like `TestSelectPeers` fails intermittently. Happens on my local machine too. I'll fix it in another PR and merge it here.

iwasaki-kenta:
Care to remove the manual binary serialization in cmd/benchmark/flood.go and cmd/wavelet/shell.go as well?

hasyimibhar:
Oh I missed those. Will do.

tx.go
```diff
@@ -69,31 +69,6 @@ func NewTransaction(creator *skademlia.Keypair, tag sys.Tag, payload []byte) Tra
 	return tx
 }
 
-func NewBatchTransaction(creator *skademlia.Keypair, tags []byte, payloads [][]byte) Transaction {
```
hasyimibhar: @iwasaki-kenta Need your thoughts on this. `NewBatchTransaction` is only used by `cmd/graph`, `cmd/benchmark`, and `tx_applier_test.go`. Now it has been replaced with `Batch` payload like this:

```go
var batch Batch
batch.AddStake(Stake{})
batch.AddTransfer(Transfer{})
batch.AddContract(Contract{})
batch.AddNop()

tx := AttachSenderToTransaction(alice, NewTransaction(alice, sys.TagBatch, batch.Marshal()))
```

I feel that it's much cleaner to do this rather than calling `NewBatchTransaction`. Now that it's no longer used anywhere, I decided to remove it.

iwasaki-kenta: That sounds much better :+1: Also, care to rebase once more with master? Some things were modified in parsing transactions.

PR #140, opened 2019-07-24
tx_json, wctl, cmd/wctl: move JSON code into wctl pkg

hasyimibhar:
Since the JSON representation of transaction payload is only used in `cmd/wctl`, it seems out of place to be put in the `wavelet` pkg.


hasyimibhar:
Closing this as it's good to standardize JSON representation of the transaction within `wavelet` package.

PR #142, opened 2019-07-24
Add tests for ledger

hasyimibhar:
Test cases:

- [x] Transfer PERL
- [x] Transfer PERL with insufficient balance
- [x] Deposit gas
- [x] Spawn smart contract
- [x] Invoke smart contract
- [x] SyncToLatestRound
- [x] Stake (place and withdraw)
- [x] ~Reward (receive and withdraw)~ (not possible as it requires a larger testnet)
- [x] SyncToLatestRound (need a way to force a node to lag a few rounds behind other nodes)
- [x] PullMissingTransactions (need a way to force a node to miss some transactions)

### DIFF
--- a/ledger_test.go
+++ b/ledger_test.go
@@ -1,11 +1,12 @@
 package wavelet
 
 import (
-	"fmt"
-	"github.com/stretchr/testify/assert"
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/perlin-network/wavelet/sys"
+	"github.com/stretchr/testify/assert"
 )
 
 // TestLedger_BroadcastNop checks that:
@@ -28,8 +29,8 @@ func TestLedger_BroadcastNop(t *testing.T) {
 	_, err := testnet.faucet.Pay(alice, 1000000)
 	assert.NoError(t, err)
 
-	// Wait for alice to receive her PERL from the faucet
-	for <-alice.WaitForConsensus() {
+	// Wait for balance to update
+	for range alice.WaitForConsensus() {
 		if alice.Balance() > 0 {
 			break
 		}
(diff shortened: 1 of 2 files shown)


ledger_test.go
```diff
@@ -0,0 +1,204 @@
...
+	assert.True(t, aliceBalance > oldBalance)
+
+	// Everyone else should see the updated balance of Alice
+	for _, node := range others {
+		assert.EqualValues(t, aliceBalance, node.BalanceOfAccount(alice))
+		assert.EqualValues(t, alice.Stake(), node.StakeOfAccount(alice))
+	}
+}
+
+func TestLedger_Reward(t *testing.T) {
```
hasyimibhar: This test takes more than 20 seconds (on my local machine) to run because it has to go through several consensus rounds to generate the rewards. Not sure if there's a better way to test the rewards generation.

ledger.go
```diff
@@ -687,7 +690,11 @@ FINALIZE_ROUNDS:
 			Uint64("round_depth", finalized.End.Depth-finalized.Start.Depth).
 			Msg("Finalized consensus round, and initialized a new round.")
 
-		//go ExportGraphDOT(finalized, l.graph)
+			//go ExportGraphDOT(finalized, l.graph)
+
+		if l.FinalizeCh != nil {
```
hasyimibhar: @iwasaki-kenta I added this as a hook to allow tests to wait for consensus round to finalize. It feels a bit like a dirty hack though. But so far it works better than just `time.Sleep`. Is there a better way to do this?

hasyimibhar: Nvm, figured out a better way without having to alter `Ledger`. Refer to #150.

iwasaki-kenta:
Let's merge these tests in - care to fix the conflict in `ledger_test.go`?

hasyimibhar:
@iwasaki-kenta Rebased, should be ready to merge.

PR #147, opened 2019-07-29
[WIP] ledger, accounts, cmd/wavelet: add cleanup

hasyimibhar:
Implement cleanup for `GC`, `Ledger`, and `wavelet/cmd`. Currently, tests for `wavelet/cmd` are failing intermittently due to lack of proper cleanup in between test cases (refer to #146). This might also be affecting tests in #142.

The cleanup is done by adding `stop` channel. Long-running goroutines (i.e. those running infinite for loop) listens for this channel. If this channel is closed, all goroutines will stop.

Might need to wait for #134 to be merged first.

TODO:

- [ ] Double check to make sure `SyncToLatestRound` behavior remains unchanged (especially considering we don't have test for `Ledger` in master yet)
- [ ] Refactor? `SyncToLatestRound` seems a bit messy, as there are multiple nested for loops. After adding `for { select { } }`, it's hard to follow through the code.

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -20,14 +20,14 @@
 package wavelet
 
 import (
-	"context"
-	"github.com/perlin-network/wavelet/avl"
-	"github.com/perlin-network/wavelet/store"
-	"github.com/pkg/errors"
 	"sync"
 	"sync/atomic"
 	"time"
 	"unsafe"
+
+	"github.com/perlin-network/wavelet/avl"
+	"github.com/perlin-network/wavelet/store"
+	"github.com/pkg/errors"
 )
 
 type Accounts struct {
@@ -44,15 +44,19 @@ func NewAccounts(kv store.KV) *Accounts {
 }
 
 // GC periodically garbage collects every 5 seconds. Only one
-// instance of GC worker can run at any time.
-func (a *Accounts) GC(ctx context.Context) {
+// instance of GC worker can run at any time. It accepts a
+// read-only channel that is used to signal the GC to stop.
+func (a *Accounts) GC(stop <-chan struct{}, stopWG *sync.WaitGroup) {
+	defer stopWG.Done()
+
 	timer := time.NewTicker(5 * time.Second)
 	defer timer.Stop()
 
 	for {
 		select {
-		case <-ctx.Done():
+		case <-stop:
 			return
+
 		case <-timer.C:
 			p := atomic.SwapPointer((*unsafe.Pointer)(unsafe.Pointer(&a.profile)), nil)
 			if p != nil {
--- a/ledger.go
+++ b/ledger.go
@@ -25,6 +25,11 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"math/rand"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/perlin-network/noise"
 	"github.com/perlin-network/noise/skademlia"
 	"github.com/perlin-network/wavelet/avl"
@@ -36,10 +41,6 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/peer"
-	"math/rand"
-	"strings"
-	"sync"
-	"time"
 )
 
 type Ledger struct {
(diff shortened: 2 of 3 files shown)


ledger.go
```diff
@@ -692,459 +736,486 @@ FINALIZE_ROUNDS:
...
 	for {
-		for {
-			conns, err := SelectPeers(l.client.ClosestPeers(), sys.SnowballK)
-			if err != nil {
+		select {
+		case <-l.stop:
+			shutdown()
+			return
+
+		default:
```
a-urth: maybe leave default branch empty, so code will fall through and there will be no need in additional nesting level?

accounts.go
```diff
@@ -44,15 +44,19 @@ func NewAccounts(kv store.KV) *Accounts {
 }
 
 // GC periodically garbage collects every 5 seconds. Only one
-// instance of GC worker can run at any time.
-func (a *Accounts) GC(ctx context.Context) {
+// instance of GC worker can run at any time. It accepts a
+// read-only channel that is used to signal the GC to stop.
+func (a *Accounts) GC(stop <-chan struct{}, stopWG *sync.WaitGroup) {
```
a-urth: how's explicit channel is better than context?
its kinda cheesy to pass workgroup here, since this method is not aware whether it will be run as goroutine or not.
IMO its better to wrap this call with another function and use workgroup there.

hasyimibhar: I replaced context with channel because it seems like `Context` is being used for cancelling the GC worker, and as far as I know, that's [not the purpose of Context](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation). I think passing a `Context` is more suitable for making a request (e.g. HTTP requests), while here, the `GC` function is meant to be run as a long-running goroutine.

As for the `WaitGroup`, I think it's safe to assume that `GC` will always be called in a goroutine due to the nature of the function (it's a GC worker), same like how we can safely assume that `CollectVotes` will always be called in a goroutine (and it also accepts a `WaitGroup` as one of its arguments).

ledger.go
hasyimibhar: Yeah that would be much better.

accounts.go
a-urth: context's `Done` is a channel with `cancel` being function which closes it. and if you treat creation of `gc` and its lifetime as a very long request then it makes sense.
i guess its all the same, but we tended to use context for cancellation purpose in most of the other parts of the project (aside `sync` channel which is used as channel because it needs to be restarted). so i'm just for having one approach for that matter ;)

hasyimibhar: I see. I wasn't aware it's already used project-wide. Then it makes senses to stick to `Context` for consistency.

---

Another way to avoid using channel explicitly for stopping the GC while still maintaining consistency is by extracting the `Ticker` ownership out of `GC`, so that it looks like this:

```go
func (a *Accounts) GC(ticker *time.Ticker, wg *sync.WaitGroup) {
    if wg != nil {
        defer wg.Done()
    }

    // No need for select
    for <-ticker.C {
        // Do GC stuff
    }
}
```

Then starting and stopping `GC` is done by interacting with the `Ticker`:

```go
// Start GC
var wg sync.WaitGroup
ticker := time.NewTicker(time.Second * 5)
wg.Add(1)
go accounts.GC(ticker, &wg)

// ...

// Stopping GC
ticker.Stop()
close(ticker.C)
wg.Wait()
```

This is synonymous to how `CollectVotes` lifetime depends on `voteChan` channel, where the goroutine is stopped by closing `voteChan` (https://github.com/perlin-network/wavelet/blob/master/ledger.go#L623-L625):

```go
workerWG.Add(1)
close(voteChan)
workerWG.Wait() // Wait for vote processor to stop
```

---

Alternatively, to be more extreme, completely extract the for loop out of `GC`:

```go
// Now GC is more open, as the caller can choose to run it as a goroutine or not.
func (a *Accounts) GC() {
    // Do GC stuff
}
```

Now `Ledger` is responsible for looping the GC, and there's no need to pass a `WaitGroup` or `Context` as `Ledger` already owns them:

```go
func (l *Ledger) RunGC() {
    ticker := time.NewTicker(time.Second * 5)
    defer ticker.Stop()

    defer l.stopWG.Done()

    for {
        select {
        case <-ticker.C:
            l.accounts.GC()

        case <-l.ctx.Done():
            return
        }
    }
}
```

hasyimibhar:
Closing this for now as it's currently not needed since #150 added `WithoutGC`, which allows creating `Ledger` without GC running for tests.

PR #149, opened 2019-08-04
Improve autocompletion indexing using a radix tree.

diamondburned:


diamondburned:
@losfair please confirm this works on Windows

PR #151, opened 2019-08-06
Preserve globals of smart contracts into account state.

losfair:


losfair:
@diamondburned The CLI changes break Wavelet on Windows so I temporarily reverted it in this PR. Can you fix it?

diamondburned:
>The CLI changes break Wavelet on Windows

How did it break? Does this deserve its own issue then?

PR #152, opened 2019-08-08
Queue transaction

hasyimibhar:
```go
paylod := wavelet.Transfer{to.PublicKey(), 1337}
ledger.QueueTransaction(wavelet.NewTransaction(keys, sys.TagTransfer, payload.Marshal()))
```

Changes:
- `AddTransaction` is replaced with `QueueTransaction`, which is only used for adding new transaction by the sender into the ledger
- Don't export `AddTransaction` and `BroadcastNop`, as they are only meant to be used internally within Wavelet (e.g. CLI and API should use `QueueTransaction` instead)
- Queue transactions are added to the ledger in the `FinalizeRounds` goroutine

TODO:

- [x] `QueueTransaction` should wait until the transaction is added to the ledger (e.g. CLI and API)
- [x] Check if queue is full
- [x] Turn errors returned by `QueueTransaction` into constants
- [x] Add some tests from #142 just to be safe
- [x] Check no tx is ignored or rejected in test

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -70,6 +70,8 @@ type Ledger struct {
 	cacheChunks   *LRU
 
 	sendQuota chan struct{}
+
+	txQueue chan Transaction
 }
 
 type config struct {
@@ -160,6 +162,8 @@ func NewLedger(kv store.KV, client *skademlia.Client, opts ...Option) *Ledger {
 		cacheChunks:   NewLRU(1024), // In total, it will take up 1024 * 4MB.
 
 		sendQuota: make(chan struct{}, 2000),
+
+		txQueue: make(chan Transaction, 1024),
 	}
 
 	ledger.PerformConsensus()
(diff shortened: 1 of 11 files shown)


ledger.go
```diff
@@ -457,6 +472,23 @@ func (l *Ledger) FinalizeRounds() {
...
+			case tx := <-l.txQueue:
+				tx = AttachSenderToTransaction(
+					l.client.Keys(), tx, l.graph.FindEligibleParents()...)
+
+				if err := l.addTransaction(tx); err != nil {
+					fmt.Println("failed to add queued transaction")
+				}
+
+			default:
+				txAvailable = false
```
iwasaki-kenta: Can just break the for loop here rather than use a `txAvailable` bool.

hasyimibhar: Really? `go vet` is saying that the code after that is unreachable. I think using `break` will break out of the `case` statement, not the `for` loop.

a-urth:
sorry for jumping in but just curious
- whats the main goal of this feature?
- won't adding transaction during finalization loop interfere/affect finalization process? if yes, then how exactly?

hasyimibhar:
@a-urth This was proposed by @iwasaki-kenta to fix the issue where tx are being ignored when one node spams the network (refer to #153). We speculated that it was happening because tx are being added to the ledger from multiple goroutines, which is causing the ledger to pick the wrong parents for each tx.

However, along the way we discovered that the issue is actually because tx are being added too fast into the ledger. There is another fix being worked on which requires the gossip mechanism to be changed.

So now I'm not sure whether this PR is still required, since it no longer achieves it's goal of fixing the ignored tx issue.

hasyimibhar:
Closing this as it's no longer needed as discussed.

Issue #160, opened 2019-08-15
Contract call fails with `insufficient gas` even though contract has sufficient gas deposited


ryanswart:
When I call a contract function that will take 700k PERL gas, on a contract that has 100mil PERL gas deposited, from a wallet with 500k PERLs, with a 1 PERL gas limit, I get an "insufficient gas" error, the wallet pays 2 PERLS in transaction fees, and the contract's gas deposit balance stays the same.

When I call the same function from a wallet with 100mil PERLs, it executes successfully, deducts gas from the contract's gas balance, and only deducts the 2 PERL transaction fee from the wallet's balance.

It is expected that when the contract pays the gas fees, the calling wallet account does not have to have enough PERLs to theoretically pay the gas fees. This allows users to run expensive contracts without needing a large amount of PERLs in their wallet.

iwasaki-kenta:
Issue was not able to be reproduced by @hasyimibhar and @a-urth. Closing until we are able to reproduce the issue.

PR #174, opened 2019-08-20
[WIP] Fix insufficient gas error when calling smart contract with sufficient gas deposit

hasyimibhar:
Refer to #160.

- [ ] Get the test case running correctly
- [ ] Find a way to extract the error returned by life vm so that the test can actually verify it's working
- [ ] Fix the error (could be something to do with incorrect gas payer)


hasyimibhar:
The issue I'm having right now is that I couldn't setup the test case to replicate the issue. According to @ryanswart, calling the `track` method with 0 amount and 1 gas limit, while the caller has 1mil PERL should succeed, but it always fails in the test case for some reason. And it ends up using more than 2 PERL as transaction fee.

ledger_test.go
```diff
@@ -124,3 +124,39 @@ func TestLedger_AddTransaction(t *testing.T) {
...
+	testnet.WaitForLatestConsensus(t)
+
+	// Deposit some gas to both smart contracts
+	_, err = testnet.faucet.DepositGas(registry.ID, 100000000)
+	assert.NoError(t, err)
+	_, err = testnet.faucet.DepositGas(contract.ID, 100000000)
+	assert.NoError(t, err)
+	testnet.WaitForLatestConsensus(t)
+
+	_, err = bob.CallContract(registry.ID, 0, 1, "track", contract.ID[:])
```
hasyimibhar: The issue I'm having right now is that I couldn't setup the test case to replicate the issue. According to @ryanswart, calling the track method with 0 amount and 1 gas limit, while the caller has 1mil PERL should succeed, but it always fails in the test case for some reason. And it ends up using more than 2 PERL as transaction fee.

a-urth:
It would be really cool to merge into master everything from this PR aside test itself.
That gonna help a lot to create other smart contracts' related tests.
@hasyimibhar @rkeene what do you guys think? maybe create another pr with all the helper functions and merge it separately?

hasyimibhar:
Sounds good to me. I can add the consensus-related helper methods to #142, which should be ready for merging anytime.

testutil.go
```diff
@@ -337,3 +500,20 @@ func loadKeys(t testing.TB, wallet string) *skademlia.Keypair {
 
 	return keys
 }
+
+func waitFor(t testing.TB, err string, fn func() bool) {
```
a-urth: this seems to be not used anywhere

testutil.go
```diff
@@ -53,9 +64,86 @@ func (n *TestNetwork) AddNode(t testing.TB, startingBalance uint64) *TestLedger
...
+		t.Fatal("timed out waiting for round")
+
+	case <-done:
+		return
+	}
+}
+
 func (n *TestNetwork) WaitForConsensus(t testing.TB) {
-	all := append(n.nodes, n.faucet)
-	TestWaitForConsensus(t, time.Second*30, all)
```
a-urth: that was the only usage of `TestWaitForConsensus`, so its not used anywhere either now

Issue #175, opened 2019-08-20
Investigate WebSocket broadcasting not working


rkeene:
It has been reported that wavelet#166 caused the websocket polling to be impaired

a-urth:
Can confirm that it stopped working in #166 
Going to find out whats happened

PR #176, opened 2019-08-20
Fix logging in collapse transaction.

a-urth:


a-urth:
Failing test is fixed here (https://github.com/perlin-network/wavelet/pull/167) btw

a-urth:
#175

PR #177, opened 2019-08-20
Memory-efficient syncing

hasyimibhar:
Syncing rounds currently requires the node to hold the entire diff in memory (both for sending and receiving). If the diff happens to be really big, it can easily overwhelm the node and taking it down. The objective of this PR to *bound* the memory usage of ledger during state syncing by writing to disk when necessary, while at the same time not sacrificing too much performance.

Major changes:
- State diff dump is now preceeded with `<no. of diffs: 8 bytes><diff id: 16 bytes>...`. This is done to allow the receiver to get the diff ids first without having to deserialize all the nodes.

- Diff is dumped into a buffer that is bounded in memory, but unbounded on disk.

- Once the diff is chunked, it's stored in a hash map which spills to on-disk KV if it exceeds the cache size.

- Downloaded chunks are written to a buffer that is bounded in memory, but unbounded on disk.

- Once all chunks are downloaded, it is deserialized into diff nodes one by one by `populateDiffs` instead of having to preload all the nodes first.


avl/tree.go
```diff
@@ -438,7 +461,11 @@ func (t *Tree) ApplyDiffWithUpdateNotifier(diff []byte, updateNotifier func(key,
 	return nil
 }
 
-func (t *Tree) ApplyDiff(diff []byte) error {
+func (t *Tree) ApplyDiffFromBytes(diff []byte) error {
```
a-urth: same goes here

avl/tree.go
```diff
@@ -376,14 +378,27 @@ func (t *Tree) iterateDiff(prevViewID uint64, callback func(n *node) bool) {
 	}
 }
 
-func (t *Tree) DumpDiff(prevViewID uint64) []byte {
+// DumpDiffAll is a wrapper of DumpDiff which simply
+// dumps the diff into memory (similar to ioutil.ReadAll).
+func (t *Tree) DumpDiffAll(prevViewID uint64) []byte {
```
a-urth: maybe `DumpDiffFull` or `DumpWholeDiff` or something similar
i got the reference to `ReadAll` but DiffAll just doesn't make much sense

avl/tree.go
```diff
@@ -396,18 +411,24 @@ func (t *Tree) IterateLeafDiff(prevViewID uint64, callback func(key, value []byt
 	})
 }
 
-func (t *Tree) ApplyDiffWithUpdateNotifier(diff []byte, updateNotifier func(key, value []byte)) error {
-	reader := bytes.NewReader(diff)
+func (t *Tree) ApplyDiffFromBytesWithUpdateNotifier(diff []byte, updateNotifier func(key, value []byte)) error {
```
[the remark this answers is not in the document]
hasyimibhar: Yeah I just added it for testing. Will clean it up later.

hasyimibhar:
From a single run:

- Badger with 500MB diff:
    - SSD: 9s (`accounts.Commit` took 700ms)
    - HDD: 51s (`accounts.Commit` took 7.2s)

- LevelDB with 500MB diff:
    - SSD: 9.9s (`accounts.Commit` took 2.2s)
    - HDD: 1m20s (`accounts.Commit` took 40s)

hasyimibhar:
Badger DB implementation is removed from this PR. Merge #217 instead to get Badger db.

Issue #155, opened 2019-08-12
Tests for checking how well external contract calls work


rkeene:
From within a smart contract, calling another smart contract may be broken -- it was working at one point, but tests need to be added around it

PR #183, opened 2019-08-23
Add subcontract invocation test.

a-urth:
Added two smart contracts (actually in another PR by mistake), one just prints (logs) hello, another one invokes smart contract by given id and function name.

Wanted to add test for balance to be sure that both smart contracts got executed, but for some reason its always different amount of gas taken.


a-urth:
#155

a-urth:
`CallContract` and `SpawnContract` are copied from #174

a-urth:
Renamed `dummy` to `hello_world`.
`rustfmt`ed both smart contracts used in test

PR #186, opened 2019-08-27
resolver: write heap pprof to disk upon oom

rkeene:
When the stall detector detects excessive memory usage, also write a binary heap dump

### DIFF
--- a/recovery.go
+++ b/recovery.go
@@ -109,14 +109,14 @@ LOOP:
 					var memStats runtime.MemStats
 					runtime.ReadMemStats(&memStats)
 
+logger.Debug().Msgf("memStats.Alloc = %v, 1048576*d.config.MaxMemoryMB = %v", memStats.Alloc, 1048576*d.config.MaxMemoryMB)
+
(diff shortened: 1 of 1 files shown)


recovery.go
```diff
@@ -114,6 +116,17 @@ LOOP:
 						_ = pprof.Lookup("heap").WriteTo(logger, 2)
 						_ = pprof.Lookup("goroutine").WriteTo(logger, 2)
 
+						crashDir := "./crashes"
+						if err := os.MkdirAll(crashDir, 0700); err == nil {
```
iwasaki-kenta: If the folder already exists, won't an error be thrown?

recovery.go
```diff
@@ -114,6 +116,17 @@ LOOP:
...
+							crashTimestamp := time.Now().Format("2006-01-02-15-04")
+							crashFile := fmt.Sprintf("%s/heap-%s.pprof", crashDir, crashTimestamp)
+
+							if heapFile, err := os.Create(crashFile); err == nil {
+								defer heapFile.Close()
```
iwasaki-kenta: Rather than defer, we should manually close right after writing to the heap file.

PR #187, opened 2019-08-28
Apply contracts to vm in same order as they came in.

a-urth:


a-urth:
@rkeene i changed names for internal structures of context and added test to ensure that it fixes this issue.

rkeene:
Awesome !

PR #195, opened 2019-08-30
Keep example updated: change result type

bofeiw:
Now the code does not compile. It compiles when i change all the return type from -> Result<(), Box<dyn Error>> to -> Result<(), String>


bofeiw:
## change result type
This makes lib.rs compile.

## add the missing argument
There is one missing argument in $contract.call which makes Vue crash. After adding the missing argument, it works.
https://github.com/johnhorsema/wavelet-todo/pull/1

iwasaki-kenta:
Thanks a lot for updating the tutorial! LGTM :).

Issue #190, opened 2019-08-29
Allow for node-local params to be set dynamically.


iwasaki-kenta:
Parameters such as:

- SNOWBALL_K
- SNOWBALL_ALPHA
- SNOWBALL_BETA
- All network-related timeouts
- SYNC_CHUNK_SIZE
- MAX_DOWNLOAD_DEPTH_DIFF
- SYNC_IF_ROUNDS_DIFFER_BY
- PRUNING_LIMIT

... should be configurable in real-time via CLI.

a-urth:
Updating pruning limit might be not that easy, since it affects buffer size and logic within `Rounds` object.

PR #196, opened 2019-08-30
Allow for node-local params to be set dynamically.

a-urth:
#190


cmd/wavelet/cli.go
```diff
@@ -141,6 +142,62 @@ func NewCLI(client *skademlia.Client, ledger *wavelet.Ledger, keys *skademlia.Ke
 			Aliases: []string{"quit", ":q"},
 			Action:  a(c.exit),
 		},
+		{
+			Name:   "update-parameters",
```
iwasaki-kenta: Update cmd to `update-param` w/ alias `up <key> <val>`.

cmd/wavelet/cli.go
```diff
@@ -141,6 +142,62 @@ func NewCLI(client *skademlia.Client, ledger *wavelet.Ledger, keys *skademlia.Ke
...
+			Action: a(c.updateParameters),
+			Flags: []cli.Flag{
+				cli.IntFlag{
+					Name:  "snowballK",
```
iwasaki-kenta: Rather than camel-case, best have them dotted-paths. For example, `snowballK` is `snowball.k`.

conf/conf.go
```diff
@@ -0,0 +1,209 @@
...
+
+func GetPruningLimit() uint8 {
+	l.RLock()
+	t := c.pruningLimit
+	l.RUnlock()
+
+	return t
+}
+
+func UpdateConfig(options ...ConfigOption) {
```
iwasaki-kenta: Best rename to `Update`, as the package name is already `conf`. Same for `ConfigOption`, being `Option`.

cmd/wavelet/cli.go
a-urth: ok, but it will be `update-params` since it allows to update more than one parameter at a time.
if no value provided for any parameter, default one will be used. thats because values to cli passed once on it's creation and then reused during each command enter.
its maybe not really convenient but its the simplest approach.

a-urth:
@iwasaki-kenta udpated and left comment

a-urth:
@rkeene @hasyimibhar @diamondburned guys, please, review

snowball.go
```diff
@@ -61,7 +52,6 @@ type Snowball struct {
 
 func NewSnowball(opts ...SnowballOption) *Snowball {
 	s := &Snowball{
-		beta:       SnowballDefaultBeta,
```
rkeene: How often do we change the Beta for one Snowball instance but not every instance tied to the configuration ?  I see we do it in the test, and it requires some hackery to work around, but is that the only case where we do care ?

Maybe we can pass in a pointer to our ledger's `conf` object instead, to pull out that Beta ?

a-urth: not sure which hackery you're about)
snowballBeta should be changed once for all node's components. but since its used only in snowball instance, its relevant only for particular snowball entity, so it doesn't really matter. its used here in such way, because snowball is created once per node lifetime, so it can't be passed as parameter or something.
I'm not sure whats the usecase for changing those values in general, i guess its just to make testing network configurations easier. so i don't think its ever will be used in "production"
as for pointer - basically current implementation is a wrapper around pointer.

PR #198, opened 2019-08-31
docs: add tutorial on adapting wasm apps to run on wavelet

Tsoyuzhu:


a-urth:
I'm not sure that we need to have tutorial which is exactly the same as post written by Ryan, since there is already descent blog post.
But, I'm sure that structure and wording should be changed because currently its really out of place, obviously, since its a blog post, not a tutorial document.

site/docs/examples/wasm-to-distributed-program.md
```diff
@@ -0,0 +1,376 @@
+---
+id: wasm-to-decentralized-program
+title: Turning Existing WebAssembly Applications Into Distributed Programs
+sidebar_label: Turning Existing WebAssembly Applications Into Distributed Programs
+---
+
+**Or, how to make an eternal Game of Life Simulator.**
```
ryanswart: We can remove this

site/docs/examples/wasm-to-distributed-program.md
```diff
@@ -0,0 +1,376 @@
...
+
+![alt text](https://miro.medium.com/max/1510/1*AXw36HaBjYVQeuERis-U6A.png "Decentralized Game of Life built on Wavelet")
+
+One of the benefits of WebAssembly that we keep going on about is the ability to re-use existing code. 
+
+There have been a couple of articles demonstrating how easy it is to build WebAssembly applications that run on the decentralized virtual machine/smart contract platform Wavelet, but it hasn’t been shown how easy it is to port existing applications yet.
```
ryanswart: Remove

site/docs/examples/wasm-to-distributed-program.md
```diff
@@ -0,0 +1,376 @@
...
+id: wasm-to-decentralized-program
+title: Turning Existing WebAssembly Applications Into Distributed Programs
+sidebar_label: Turning Existing WebAssembly Applications Into Distributed Programs
+---
+
+**Or, how to make an eternal Game of Life Simulator.**
+
+![alt text](https://miro.medium.com/max/1510/1*AXw36HaBjYVQeuERis-U6A.png "Decentralized Game of Life built on Wavelet")
+
+One of the benefits of WebAssembly that we keep going on about is the ability to re-use existing code.
```
ryanswart: Change to "This tutorial aims to show how easy it is to re-use Rust application code in Wavelet."

ryanswart:
Hi @a-urth , main reason for having the tutorial here is so that it can be kept up to date with wavelet changes; its a pain to do so on medium as you have to have a separate gist for each code block. Also it provides a central place for developers to find tutorials, which they may have trouble finding otherwise.

Also, as the markdown is the source for the blog post, keeping it the same as the blog post theoretically makes it easier to update the blog post as well, however I agree that it would be better to re-word it if we don't want to keep the blog post up to date.

Will add some comments for possible changes to make it flow better as documentation.

a-urth:
@ryanswart awesome, thanks for answer ;)

Issue #185, opened 2019-08-26
Connect/disconnect/restart API and CLI methods.


rkeene:
Create Disconnect and Connect API and CLI Methods.

CLI: `/connect <host> [<port>]`
CLI: `/disconnect <host> [<port>]`
CLI: `/restart <--hard>`
API: `/node/connect/<:host>[/<:port>]`
API: `/node/disconnect/<:host>[/<:port>]`
API: `/node/restart?hard=true`

For Connect calls: port missing, use configured port
For Disconnect calls: port missing, match any port (disconnect all if multiple matches)

iwasaki-kenta:
Finished `connect` and `disconnect` commands, and pushed directly to `master`.

iwasaki-kenta:
Assigned @a-urth. For `restart`, passing in a `--hard` flag either in the CLI or via the HTTP API will flush/delete the node's persisted database on-disk.

The node will be restarted via the method specified in recovery.go, where the node is first shut down then restarted.

a-urth:
From issue description it seems that `port` should be optional parameter, but in `cli` its implemented as mandatory.
@rkeene @iwasaki-kenta so should it be optional? and if yes, what value should be used as default?

iwasaki-kenta:
> From issue description it seems that `port` should be optional parameter, but in `cli` its implemented as mandatory.
> @rkeene @iwasaki-kenta so should it be optional? and if yes, what value should be used as default?

Port should be made mandatory for `connect`. However for `disconnect`, we should allow for the specification of something like a regex to match addresses.

a-urth:
also, in examples @rkeene marked host and post as url parts, but i think its better to make those first of all as `POST` request, then make host+port a single value and move it to json payload. so it will be consistent with `cli` and there will be no problems with special characters when we'll add regexp for disconnect.
what do you think guys @rkeene @iwasaki-kenta ?

PR #204, opened 2019-09-02
Add api endpoint for connecting disconnecting node.

a-urth:
#185

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -34,14 +34,13 @@ import (
 type Accounts struct {
 	sync.RWMutex
 
-	kv   store.KV
 	tree *avl.Tree
 
 	profile *avl.GCProfile
 }
 
 func NewAccounts(kv store.KV) *Accounts {
-	return &Accounts{kv: kv, tree: avl.New(kv)}
+	return &Accounts{tree: avl.New(kv)}
 }
 
 // GC periodically garbage collects every 5 seconds. Only one
--- a/api/middleware.go
+++ b/api/middleware.go
@@ -113,12 +113,7 @@ func oAuth2(ctx *fasthttp.RequestCtx) string {
 
 func (g *Gateway) auth(next fasthttp.RequestHandler) fasthttp.RequestHandler {
 	return func(ctx *fasthttp.RequestCtx) {
-		if len(conf.GetSecret()) == 0 {
-			next(ctx)
-			return
-		}
-
-		if oAuth2(ctx) == conf.GetSecret() {
+		if len(conf.GetSecret()) > 0 && oAuth2(ctx) == conf.GetSecret() {
 			next(ctx)
 			return
 		}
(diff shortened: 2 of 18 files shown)


a-urth:
Added `restart` cli command and corresponding endpoint with `hard` flag which wipes db if its present.

rkeene:
I think this needs some sort of authentication system built first (or as part of this changeset, even) so that anonymous users on the Internet don't wipe our testnet (or even disconnect/connect random nodes to our cluster).

a-urth:
@rkeene i think api is designed to be used from lens only, if node won't be exposed to external world (by http) this should not be a problem

rkeene:
Currently, on the testnet we do expose the HTTP API to the world and nothing in this PR seems to restrict it as far as I can tell...

a-urth:
@rkeene i don't think that its possible to connect to testnet outside of lens or clicker as of now
@iwasaki-kenta ? and your opinion an auth for api in general

rkeene:
https://testnet.perlin.net/   leads directly to the HTTP API for the testnet.

For example:

```
$ curl -sSL https://testnet.perlin.net/ledger | jq --tab .
{
        "public_key": "0f569c84d434fb0ca682c733176f7c0c2d853fce04d95ae131d2f9b4124d93d8",
        "address": "10.0.13.124:3000",
        "num_accounts": 1286,
        "round": {
                "merkle_root": "87cfefa37b967555890f0366d29e93bc",
                "start_id": "41586bd3a7fd5a5881393304dedbea16f6d9b2d7c83e24fcd9b3d8bda63e5dd9",
                "end_id": "2cf5ed86b3fcb1e02a6ad23d222c55052e689222b291a27d90e35573a6964c01",
                "applied": 8,
                "depth": 8,
                "difficulty": 8
        },
        "peers": [
                {
                        "address": "10.0.13.121:3002",
                        "public_key": "ba41cbb19074d27727607992e123345dd82f1fc115233b5ae43a07c00ddda3df"
                },
                {
                        "address": "10.0.13.123:3001",
                        "public_key": "487cbbbf68fc15d67a1a074951172b4cfd043489e93d7ae71c8e4aa0be1593b0"
                }
        ]
}
```

iwasaki-kenta:
@rkeene Can you leave feedback on what the ideal authentication system would be given the Docker Swarm setup?

rkeene:
For the Docker Swarm setup it would be ideal if there were a simple shared secret that were used.

The simplest implementation I can think of would be a CLI flag `--api.secret` (and accompanying environment variable) containing to specify that value, or if not specified either those APIs are disabled or a random 256-bit hex-encoded value is used (and displayed to the user, similar to the node's private key -- it could even *BE* the node's private key).

To use it, each request would need to include it as a parameter:

```json
{
        "api_secret": "<sharedSecret>",
        "host": "<host>",
        "port": <port>
}
```

PR #208, opened 2019-09-05
Add support to enable HTTPS on the API server

AhmadMuzakkir:
Implement HTTPS API server using Let's Encrypt.

Add new flag `-api.host` which is the domain. 

To enable HTTPS, pass the flag `-api.host`, and also `api.port`.

The `api.port` is used to start a HTTP server to handle Let's Encrypt auth callbacks, and redirect all  HTTP traffic into HTTPS.

### DIFF
--- a/api/mod.go
+++ b/api/mod.go
@@ -32,6 +32,7 @@ import (
 	"github.com/perlin-network/wavelet/sys"
 	"github.com/pkg/errors"
 	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttpadaptor"
 	"github.com/valyala/fasthttp/pprofhandler"
 	"github.com/valyala/fastjson"
 	"golang.org/x/crypto/acme/autocert"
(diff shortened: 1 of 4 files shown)


rkeene:
Let's make this only use the method `tls-alpn-01`, since `tls-sni-01` is deprecated, and `http-01` won't work without further implementation work (and even then would be sub-optimal).

rkeene:
2 more changes:

1. We do need the plain old HTTP API enabled after all, even if we have HTTPS enabled -- because HTTPS could be not working and we need to manage the node
2. The ability to specify where autocert puts those certs.  Right now it's always `./certs/`, but it'd be better if we can specify a path

rkeene:
Also, it looks like HTTP/2.0 support is broken -- we should disable it for now:

```
$ curl --http1.1 https://wavelet.host
Not Found
$ curl -v https://wavelet.host
...
http2 error: Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly.
...
curl: (16) Error in the HTTP2 framing layer
$
```

rkeene:
Failing tests are not related to this change, merging in

PR #214, opened 2019-09-09
Fix a panic (send on closed channel) in SyncToLatestRound

AhmadMuzakkir:
The panic is caused by a single character mistake in ledger.SyncToLatestRound()
The mistake cause a new variable to be created instead of updating the existing variable.

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -1002,7 +1002,7 @@ func (l *Ledger) SyncToLatestRound() {
 
 		restart := func() { // Respawn all previously stopped workers.
 			snowballK := conf.GetSnowballK()
-			syncVotes := make(chan vote, snowballK)
+			syncVotes = make(chan vote, snowballK)
 			go CollectVotes(l.accounts, l.syncer, syncVotes, voteWG, snowballK)
 
 			l.sync = make(chan struct{})


AhmadMuzakkir:
> thats not a mistake since on line 996 this channel is supposed to be closed.

Yes. But I think it's supposed to reinitialised in the restart func.

a-urth:
it will be reinitialized anyway, variable just will be reused from outer scope of the func. otherwise you will get panic in shutdown about closing already closed channel

AhmadMuzakkir:
Are saying the syncVotes channel should declared inside the loop ?

Currently, the syncVotes channel is declared outside the loop. That's what causing the panic when it's closed but the start the loop again.

a-urth:
It should be defined outside of the loop, this way it will be closed and created a new again.
Previously it was defined as field of `Ledger` structure, but i changed that since it was unnecessary. If you want you can change it back again, but it should be defined anyway outside of sync loop.

a-urth:
ok, wait, i've mixed your changes. @AhmadMuzakkir you fixed it correctly, sorry for misleading.

ledger.go
```diff
@@ -1002,7 +1002,7 @@ func (l *Ledger) SyncToLatestRound() {
 
 		restart := func() { // Respawn all previously stopped workers.
 			snowballK := conf.GetSnowballK()
-			syncVotes := make(chan vote, snowballK)
+			syncVotes = make(chan vote, snowballK)
```
a-urth: good catch. i managed to miss it even second time :D

Issue #200, opened 2019-09-02
Improve safety and liveness of Snowball based on DAG structure.


iwasaki-kenta:
To further mitigate safety/liveness issues by the way, there are more changes I, @rkeene and @losfair propose to append to Snowball that may only apply w.r.t. Wavelet's DAG and consensus round structure.

One huge improvement is to allow for nodes to overwrite their preferred round decision based on the contents of said round.

A node may choose to overwrite their preferred round if:

Some round that is valid and collapsible has an ending critical transaction that is at a lower depth than the node's current preferred round's ending critical transaction depth.
Some round that is valid and collapsible encompasses a larger number of ancestral transactions from it's round's ending critical transaction.

a-urth:
So IMO this is related more to votes counting than to snowball per se.
But question I got is how should we exactly treat less depth and bigger number of transactions comparing to stakes? I mean those values either should be weighted or we should consider them in some order - e.g. stake first, then depth, then number of transaction.
Its clear that we should prefer round which has end transaction at lower depth or more transactions than current. But that will happen after snowball, since we give snowball some round which we determine as preferred by majority by some characteristic. Currently its only stake. So the question is how should we find out majority based on 3 criteria instead of one?

iwasaki-kenta:
> So IMO this is related more to votes counting than to snowball per se.
> But question I got is how should we exactly treat less depth and bigger number of transactions comparing to stakes? I mean those values either should be weighted or we should consider them in some order - e.g. stake first, then depth, then number of transaction.
> Its clear that we should prefer round which has end transaction at lower depth or more transactions than current. But that will happen after snowball, since we give snowball some round which we determine as preferred by majority by some characteristic. Currently its only stake. So the question is how should we find out majority based on 3 criteria instead of one?

Have them all as weights and multiply them altogether for now.

You should however leave the option to allow potentially to weigh a certain set of weights more than others (e.g. stake should be taken more into consideration vs depth vs number of transactions).

PR #218, opened 2019-09-10
Change vote counting for finalization process.

a-urth:
#200 

- separate vote counting goroutines for syncing and finalization
- use transactions num and round depth in calculating majority for finalization voting
- add tests for finalization voting
- fix api test for ledger status


rkeene:
I tested this with 3 nodes and benchmarking running against all 3 nodes and it seems to resolve the issue

Issue #230, opened 2019-09-19
Syncing Under Load Fails


rkeene:
Investigation in progress

a-urth:
First of all I'm not even sure that we should be trying cases with spamming nodes while preventing consensus from happening (having less nodes running than necessary for consensus) because that never should a case in production. @iwasaki-kenta thoughts?
But nonetheless, I tried couple of different setups and for branch with changes in vote counting setup described above is actually resolves.
I tried running 3 and 4 nodes network while benchmarking 2,3 and 3 nodes for respective setup. Restarting one of the nodes always results into target node getting back into finalization process even without stopping benchmark on any of the nodes in the network.
I should say that such test is really heavy on CPU and all potential problems might be because of nodes are not getting enough of resources to correctly run on their own.

rkeene:
What about not updating the graph if there are less than `WAVELET_K` peers ?

rkeene:
I tested with the `vote-countint-for-finalize` branch and disallowing `ledger.AddTransaction()` (diff below) if there are less than K peers (test configuration: `WAVELET_K=2`, `WAVELET_NODES=3`, stake for each node was equal at 1000 kens)

```diff
--- a/ledger.go
+++ b/ledger.go
@@ -218,6 +218,10 @@ func NewLedger(kv store.KV, client *skademlia.Client, opts ...Option) *Ledger {
 // is returned if the transaction has already existed int he ledgers graph
 // beforehand.
 func (l *Ledger) AddTransaction(tx Transaction) error {
+       if len(l.client.ClosestPeerIDs()) < conf.GetSnowballK() {
+               return errors.New("Unable to accept transactions with less than K peers")
+       }
+
        err := l.graph.AddTransaction(tx)

        if err != nil && errors.Cause(err) != ErrAlreadyExists {
```

After 60 rounds I restarted node 1 and let it try to resync.  These were the results:

[...]

PR #235, opened 2019-09-24
[WIP] Fix tx with depth less than root depth + DEPTH_DIFF still marked as missing

hasyimibhar:
Attempt to diagnose issue #230

### DIFF
--- a/graph.go
+++ b/graph.go
@@ -22,9 +22,10 @@ package wavelet
 import (
 	"bytes"
 	"encoding/hex"
-	"github.com/perlin-network/wavelet/conf"
 	"sort"
 	"sync"
+
+	"github.com/perlin-network/wavelet/conf"
 
 	"github.com/google/btree"
 	"github.com/perlin-network/noise/edwards25519"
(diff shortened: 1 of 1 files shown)


graph.go
```diff
@@ -448,7 +449,10 @@ func (g *Graph) Missing() []TransactionID {
 	g.RLock()
 	missing := make([]TransactionID, 0, len(g.missing))
 
-	for id := range g.missing {
+	for id, depth := range g.missing {
+		if g.rootDepth > conf.GetMaxDepthDiff()+depth {
```
a-urth: but how is this possible? such transactions should be deleted from missing on root depth update

rkeene:
Tested, there was no change with this.

graph.go
rkeene: It's not clear -- it seems to have not been effective as we still got the error identified in #230 with this branch

hasyimibhar: I am not sure yet. This is not an actual fix, just trying to see if doing this makes the error go away. Looks like the error is still there.

rkeene:
Let's try something else ! 👍

Issue #233, opened 2019-09-24
Update Reward Computation Mechanism


rkeene:
We want to change the way fees are rewarded in a round to be proportional to the stake of each *sender* in the round, with numerical rounding applied.

For example a round with 5 transactions, each at 10 kens in fees:

   1. SenderA (fee 10 kens)
   2. SenderB (fee 10 kens)
   3. SenderC (fee 10 kens)
   4. SenderA (fee 10 kens)
   5. SenderD (fee 10 kens)

With the following stake weight:

  1. SenderA.stake = 10
  2. SenderB.stake = 3
  3. SenderC.stake = 5
  4. SenderD.stake = 8

The total  stake in the round is `SenderA.stake (10) + SenderB.stake (3) + SenderC.stake (5) + SenderA.stake (10) + SenderD.stake (8)` == `36`

The reward would be.
  1. For SenderA: `20 / 36` == 55.56% (27.78 kens == 28 kens)
  2. For SenderB: `3 / 36` == 8.33% (4.17 == 4 kens)
  3. For SenderC: `5 / 36` == 13.89% (6.94 kens == 7 kens)
  4. For SenderD: `8 / 36` == 22.22% (11.11 kens = 11 kens)

(It's significant to note that the rewards should be exactly equal to the fees, so rounding must be done in such a way to achieve that)

The advantage here is computing the reward (as well as the fees, added in #232) only requires walking the graph one time, and fees are distributed more evenly rather than encouraging nodes to create low-value transactions to increase their likelihood of getting a reward)

a-urth:
Btw, currently we take fee for all transactions within round, even for nops. And I remember this issue was brought up from community as well. Do we plan to change that?

Issue #232, opened 2019-09-24
Update Transaction Fee Computation


rkeene:
We want to change the way fees are computed to be charged proportional to the transaction size.
(e.g., `transaction.fee = tx.sizeInBytes * feePerByte` where `feePerByte` is a network-wide constant)

PR #239, opened 2019-09-25
Change transaction fee and reward calculations.

a-urth:
#232 #233 

Also changed naming for `collapse_test`

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -311,6 +311,15 @@ func (tx Transaction) String() string {
 	return fmt.Sprintf("Transaction{ID: %x}", tx.ID)
 }
 
+func (tx Transaction) Fee() uint64 {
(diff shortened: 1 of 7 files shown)


tx.go
```diff
@@ -311,6 +311,15 @@ func (tx Transaction) String() string {
 	return fmt.Sprintf("Transaction{ID: %x}", tx.ID)
 }
 
+func (tx Transaction) Fee() uint64 {
+	fee := uint64(sys.TransactionFeeMultiplier * float64(len(tx.Payload)))
```
iwasaki-kenta: Rather than looking at the size of the payload, we should look at the entire size of the transaction. This way, we would know by default what the transaction fee is.

PR #241, opened 2019-09-27
Increase test coverage.

hasyimibhar:
- [x] Remove generated code and test utilities from test coverage
- [x] Remove unused code
- [x] Add more tests

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -20,7 +20,12 @@
 package wavelet
 
 import (
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
 	"math/rand"
+	"strconv"
+	"strings"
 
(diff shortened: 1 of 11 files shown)


utils.go
```diff
@@ -53,59 +49,3 @@ func SelectPeers(peers []*grpc.ClientConn, amount int) ([]*grpc.ClientConn, erro
 
 	return activePeers, nil
 }
-
-func ExportGraphDOT(round *Round, graph *Graph) {
```
iwasaki-kenta: I feel we should still keep the following function in. Thoughts?

hasyimibhar: I'll revert it.

PR #243, opened 2019-09-27
Ignore tx error caused by adding transactions with too low depth

hasyimibhar:
If a transaction depth is too low due to pruning, the transaction is now ignored.


hasyimibhar:
This PR implements `Ledger.Close`. I created PR #244 specifically for that (ideally we should strive so that each PR only does 1 thing), so make sure that PR gets merged first. After it's merged, I will rebase this PR on top of master.

Issue #181, opened 2019-08-23
Wavelet's ledger needs to have proper shutdown mechanism


a-urth:
There is no proper shutdown for ledger as of now, thats why in tests we currently shut down grpc server and wait for some amount of time. Which makes tests non deterministic, since there a lot of requests still trying to access nodes and sometimes port collision happens.

hasyimibhar:
I opened PR #147 to address this issue, but I later closed it because I felt it was not needed yet since `WithoutGC` option is introduced for `NewLedger` (GC was causing tests to panic as the GC goroutine was not killed properly between tests),

But recently I faced this issue where a node takes too long to exit. It turns out the node's HTTP server was blocking on `GracefulStop` (probably because there are processes still making requests to it). I had to switch to `Stop` [to make the tests pass](https://github.com/perlin-network/wavelet/pull/177/files#diff-b41699f57b87bee7a2a750c1b079fb63R212). While it works for now, I think it's better if we have proper shutdown for the ledger.

PR #222, opened 2019-09-14
Separate cmd/wavelet into a daemon and HTTP CLI interface.

diamondburned:
- CLI now only uses wctl to interact with Wavelet
- CLI can probably (untested) connect directly to Testnet
- `server` package in `cmd/wavelet` hosts the whole Wavelet "server" now
- `wctl` API changes, now uses `On*` handlers, refer to `cmd/wavelet/events.go`
- Non-breaking API changes, now has more fields to it (**TODO: Update API docs**)
- Connect and Disconnect functions 
- `api/` tests
- Poll functions aren't all done yet, though cores are done
- API docs

PR #244, opened 2019-09-27
Cleanup ledger.

hasyimibhar:
Refer to #181. Having proper cleanup will make our tests more robust, as currently tests may sometimes fail as it's possible for different instances of `Ledger` to interact with each other between tests. This usually happens with tests in `cmd/wavelet` package.

Things to do when stopping wavelet:
- [cli] Stop API server
- [cli] Stop GRPC (close all client connections, and then the server)
- [ledger] Stop all consensus-related goroutines
- [ledger] Stop GC goroutine (if enabled)
- [ledger] Stop stall detector goroutine
- [ledger] Wait for all of the goroutines above to complete
- [cli] Close KV

TODO:
- [ ] Handle `SIGTERM` and `SIGINT`
- [ ] Handle close when restarting node via `StallDetector`

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -45,7 +45,10 @@ func NewAccounts(kv store.KV) *Accounts {
 
 // GC periodically garbage collects every 5 seconds. Only one
 // instance of GC worker can run at any time.
-func (a *Accounts) GC(ctx context.Context) {
+func (a *Accounts) GC(ctx context.Context, wg *sync.WaitGroup) {
+	wg.Add(1)
+	defer wg.Done()
+
 	timer := time.NewTicker(5 * time.Second)
 	defer timer.Stop()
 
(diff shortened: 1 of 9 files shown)


hasyimibhar:
Currently, GC is stopped using `context.Context`, while stall detect is stopped using `chan struct{}`. Waiting for goroutine finishes is done using `sync.WaitGroup`. Ideally we should stick to just one way of stopping goroutines.

hasyimibhar:
There is an issue with handling `SIGTERM` and `SIGINT`. For some reason, `readline.Instance` is blocking if `readline.Instance.Close` is called after the signal is received. However, this is not a critical issue, as according to @rkeene, `wavelet-stack` uses `kill -9` to restart node anyway. We can fix this issue in a later PR.

hasyimibhar:
As for the issue regarding `StallDetector`, it requires moving `StallDetector` out of `Ledger`. It's probably better to do it after PR #222 is merged, as it requires modifying `cmd/wavelet` and `api`.

PR #256, opened 2019-10-02
Fix captured value

rkeene:
When dealing with a captured value which is a pointer we want to make sure we are dealing with the iterated value rather than the last one


AhmadMuzakkir:
It seems the failing tests are because of the line 996.

I could not figure out why.

PR #259, opened 2019-10-02
Increase the size of the mockStdout buffer

rkeene:
As we write more and more debugging/informational logs the existing buffer may get full


rkeene:
Replaced by #260

PR #270, opened 2019-10-07
Keep the downloaded diff dump on disk

hasyimibhar:
The ledger writes the downloaded diff dump to a temporary file, which is not removed when the node is restarted. This will help to debug the issue where a node fails to sync to the latest round due to a missing AVL node in the downloaded diff dump.


rkeene:
@hasyimibhar I think this PR has been overcome-by-events and we can close it out.   Can you confirm ?

PR #275, opened 2019-10-15
Remove gossip mechanism

hasyimibhar:
Moving forward, each node will only pull transactions from other peers. Transactions will no longer be gossiped.


hasyimibhar:
Moved to #276.

PR #276, opened 2019-10-15
Remove gossip mechanism.

hasyimibhar:
Moving forward, each node will only pull transactions from other peers. Transactions will no longer be gossiped.


go.mod
```diff
@@ -13,7 +13,8 @@ require (
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
 	github.com/djherbis/buffer v1.1.0
 	github.com/fasthttp/websocket v1.4.0
-	github.com/gogo/protobuf v1.3.0
+	github.com/gogo/protobuf v1.3.1
+	github.com/golang/protobuf v1.3.2
```
iwasaki-kenta: Make sure to stick to relying on gogo/protobuf only.

PR #277, opened 2019-10-15
[WIP] Remove transaction creator and graph

AhmadMuzakkir:
As part of v2 update, we need to remove transactions creator, graph, and nop.
The graph will be replaced with mempool.

- [x] Remove transaction creator
- [ ] Remove graph
- [ ] Add mempool
- [ ] Remove nop


wctl/msg.go
```diff
@@ -187,7 +184,6 @@ func (t *Transaction) ParseJSON(v *fastjson.Value) {
 	t.Payload = v.GetStringBytes("payload")
 	t.AccountsMerkleRoot = string(v.GetStringBytes("accounts_root"))
 	t.SenderSignature = string(v.GetStringBytes("sender_signature"))
```
iwasaki-kenta: Care to rename SenderSignature to just Signature?

PR #283, opened 2019-10-21
Send list of tx id instead of bloom filter for pulling transactions

hasyimibhar:
This PR is to investigate the performance difference between:
- pulling transactions using bloom filter
- pulling transactions using naive method (the sender node will send a list of tx id it has)


iwasaki-kenta:
I'll review the PR.

protocol.go
```diff
@@ -169,14 +167,18 @@ func (p *Protocol) CheckOutOfSync(ctx context.Context, req *OutOfSyncRequest) (*
 func (p *Protocol) PullTransactions(ctx context.Context, req *TransactionPullRequest) (*TransactionPullResponse, error) {
 	res := &TransactionPullResponse{Transactions: [][]byte{}}
 
-	filter := &bloom.BloomFilter{}
-	if _, err := filter.ReadFrom(bytes.NewReader(req.Filter)); err != nil {
-		return nil, err
+	// Build a lookup table from the list of transaction IDs
+	var id TransactionID
+	lookup := map[TransactionID]bool{}
```
AhmadMuzakkir: Suggestion to use map[TransactionID]struct{} to save memory.

PR #287, opened 2019-10-22
Cleaned up locking for Transactions

diamondburned:


transactions.go
```diff
@@ -12,29 +15,50 @@ func NewTransactions() *Transactions {
 }
 
 func (t *Transactions) Add(tx *Transaction) {
+	t.mu.Lock()
 	t.buffer[tx.ID] = tx
+	t.mu.Unlock()
```
iwasaki-kenta: Prefer using `defer t.mu.Unlock()`.

transactions.go
```diff
@@ -12,29 +15,50 @@ func NewTransactions() *Transactions {
...
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
 	return t.buffer[id]
 }
 
 func (t *Transactions) Has(id TransactionID) bool {
-	_, exist := t.buffer[id]
-	return exist
+	return t.Get(id) != nil
```
iwasaki-kenta: Curious, is it cheaper to do `_, exist := t.buffer[id]; return exist` over `return t.buffer[id] != nil`?

diamondburned: They both access the map once and they'd do the same thing, I don't think so.

PR #292, opened 2019-10-23
Initial work for having client SDKs maintain local nonce and block height, and wipe away transactions from finalized blocks.

hasyimibhar:
This PR adds the following changes:
- Adds `/nonce/:id` API endpoint, which returns the latest nonce and known block height of the node
- Modifies all clients to locally increment nonce when creating transactions, and keep track of block height by polling the consensus


iwasaki-kenta:
At lunch right now, I'll be able to review it in about a hour.

api/msg.go
```diff
@@ -249,6 +249,18 @@ func (s *ledgerStatusResponse) marshalJSON(arena *fastjson.Arena) ([]byte, error
 	return o.MarshalTo(nil), nil
 }
 
+type nonceResponse struct {
```
diamondburned: This struct should export its fields to make it consistent with other structs. It should also have `json` tags for the same reason.

cmd/benchmark/main.go
```diff
@@ -215,11 +216,53 @@ func commandRemote(c *cli.Context) error {
...
+	resp, err := client.GetAccountNonce(hex.EncodeToString(id[:]))
+	if err != nil {
+		return err
+	}
+
+	nonce := resp.Nonce
+	block := resp.Block
+
+	// Poll block height changes
+	consensus, err := client.PollLoggerSink(nil, wctl.RouteWSConsensus)
```
iwasaki-kenta: For block height and nonce, care to have this done and initialised inside wctl, and remove the nonce and block height param in SendTransaction?

PR #294, opened 2019-10-23
Properly increment nonce and block height in wctl client

hasyimibhar:
This is the continuation of PR #292.


iwasaki-kenta:
Reviewing it now.

PR #296, opened 2019-10-24
Check block height when adding transaction

hasyimibhar:
When a new transaction is added to the ledger, it must be checked so that `Block` is set to the next block height that will be finalized.

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -53,9 +53,10 @@ const (
 )
 
 var (
-	ErrOutOfSync     = errors.New("Node is currently ouf of sync. Please try again later.")
-	ErrAlreadyExists = errors.New("transaction already exists in the graph")
-	ErrMissingTx     = errors.New("missing transaction")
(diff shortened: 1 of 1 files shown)


a-urth:
on it

ledger.go
```diff
@@ -239,7 +240,14 @@ func (l *Ledger) Close() {
 // is returned if the transaction has already existed int he ledgers graph
 // beforehand.
 func (l *Ledger) AddTransaction(txs ...Transaction) error {
-	l.transactions.BatchAdd(l.blocks.Latest().ID, txs...)
+	block := l.blocks.Latest()
+	for _, tx := range txs {
+		if tx.Sender == l.client.Keys().PublicKey() && tx.Block != block.Index+1 {
```
iwasaki-kenta: The only check needed is tx.Block >= latestBlock.Index - GetPruningLimit(). To prevent uint64 underflow, tx.Block + GetPruningLimit() >= latestBlock.Index.

The sender doesn't need to be checked.

PR #297, opened 2019-10-24
Add more debug info into cmd/wavelet, and make what transactions are proposed into a block more strict.

iwasaki-kenta:
api, wctl, actions: provide info about what block proposal nodes prefer
transactions: only propose transactions that are suitable for the current block height

`num_tx` in LedgerStatus is the number of transactions in the mempool, and `num_tx_in_store` is the total number of transactions archived by a node.


hasyimibhar:
I'm reviewing it.

PR #299, opened 2019-10-24
Minor cleanup over transactions, and fix off-by-one error in wctl.

iwasaki-kenta:
transactions: delete pruned transactions in the for loop over buffer instead of in amidst iterating over mempool index, and add debug log
wctl: make sure block height assigned to transaction is the latest block height, not the next block height
tx: simplify NewTransaction() to use NewSignedTransaction()

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -126,12 +126,10 @@ func (t *Transactions) ReshufflePending(next Block) int {
 
 		tx := t.buffer[item.id]
 
-		if next.Index >= tx.Block+uint64(conf.GetPruningLimit()) {
-			return true
+		if next.Index < tx.Block+uint64(conf.GetPruningLimit()) {
+			item.index = tx.ComputeIndex(next.ID)
+			items = append(items, item)
 		}
-
-		item.index = tx.ComputeIndex(next.ID)
-		items = append(items, item)
 
 		return true
 	})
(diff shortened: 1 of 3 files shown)


wctl/tx.go
```diff
@@ -83,7 +83,7 @@ func (c *Client) SendTransaction(tag byte, payload []byte) (*TxResponse, error)
 	var res TxResponse
 
 	nonce := atomic.AddUint64(&c.Nonce, 1)
-	block := atomic.LoadUint64(&c.Block) + 1
+	block := atomic.LoadUint64(&c.Block)
```
hasyimibhar: I thought when we add a transaction, block height is set to the next block that we are finalizing on, which is latest known block height + 1?

iwasaki-kenta: The block height isn't strict - its just to tell other nodes the latest block height they are on when they proposed the tx.

transactions.go
```diff
@@ -138,6 +136,8 @@ func (t *Transactions) ReshufflePending(next Block) int {
 		return true
 	})
 
+	fmt.Printf("Restoring %d items into mempool which originally had %d items, with next block height being %d.\n", len(items), t.index.Len(), next.Index)
```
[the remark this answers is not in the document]
iwasaki-kenta: This is just `printf` debugging in this case until we solve the bug of the mempool emptying out.

PR #302, opened 2019-10-24
Pull missing transactions and sync unknown transactions as separate processes

a-urth:
- pull missing transactions directly by id within one request (assumption that it always will be only a few of those);
 - sync mempool (index) transactions through grpc stream using bloom filter as source of known transactions;
 - remove error as return value from ledger's `AddTransaction` since it never returns an actual error

TODO:
 - ~~move transaction syncing parameters to config~~


iwasaki-kenta:
Do logs come up indicating that transactions pulled which were missing are not being added because of height difference, etc?

a-urth:
There are no logs for that, since this validation results in just `return` in `transactions.add`. I added prints, but didn't reproduce it yet, that why it was and assumption.

rkeene:
I'm reviewing this now

ledger.go
```diff
@@ -238,12 +243,14 @@ func (l *Ledger) Close() {
 // invalid or fails any validation checks, an error is returned. No error
```
rkeene: The comment for this function needs to be updated to indicate that it accepts many transactions and remove references to failing

ledger.go
```diff
@@ -380,56 +383,186 @@ func (l *Ledger) PullTransactions() {
...
-						responseChan <- nil
+				res, err := stream.Recv()
+				if err != nil {
+					logger.Error().Err(err).Msg("failed to receive sync transactions header")
+					return
+				}
+
+				count := res.GetTransactionsNum()
+				if count == 0 || count > maxSize {
+					return
```
rkeene: What do you think about logging the case where too many transactions are received (and thus we don't accept any?)

```suggestion
					if count != 0 {
						logger.Debug().Msg("Bad number of transactions would be received").Uint64("count", count)
					}
					return
```

ledger.go
```diff
@@ -380,56 +383,186 @@ func (l *Ledger) PullTransactions() {
...
+				}
 
+				if count > 0 {
```
rkeene: This should always be true at this point because of the check right above it

ledger.go
```diff
@@ -380,56 +383,186 @@ func (l *Ledger) PullTransactions() {
...
-					cancel()
 
-					responseChan <- batch
+					res, err := stream.Recv()
+					if err != nil {
+						logger.Error().Err(err).Msg("failed to receive sync transactions header")
+						return
+					}
+
+					transactions := res.GetTransactions()
```
rkeene: What do you think about calling this "`rawTransactions`" or "`transactionsRaw`" or "`txsRaw`" and the "`txs`" below as "`transactions`" to make it a bit more unambiguous ?

ledger.go
```diff
@@ -380,56 +383,186 @@ func (l *Ledger) PullTransactions() {
 			select {
```
rkeene: This whole `select` section can be removed since the very first thing the top of the loop does is this

ledger.go
```diff
@@ -380,56 +383,186 @@ func (l *Ledger) PullTransactions() {
...
+		case <-l.sync:
+			return
+		case <-time.After(100 * time.Millisecond):
+		}
+
+		snowballK := conf.GetSnowballK()
+
+		peers, err := SelectPeers(l.client.ClosestPeers(), snowballK)
+		if err != nil {
+			select {
```
rkeene: This whole `select` section can be removed (same as `SyncTransactions()`)

transactions.go
```diff
@@ -59,10 +59,14 @@ func (t *Transactions) BatchAdd(block BlockID, transactions ...Transaction) {
 
 func (t *Transactions) add(block BlockID, tx Transaction) {
 	if t.height >= tx.Block+uint64(conf.GetPruningLimit()) {
+		fmt.Println(">>>>>>>>>>>> why??????", t.height, tx.Block)
```
rkeene: We'll leave this debugging in for now as we track it down

transactions.go
```diff
@@ -59,10 +59,14 @@ func (t *Transactions) BatchAdd(block BlockID, transactions ...Transaction) {
...
 		return
 	}
 
 	if _, exists := t.buffer[tx.ID]; exists {
+		if _, ok := t.missing[tx.ID]; ok {
+			fmt.Println(">>>>>>>>>>>> WHOAAAAAAAAAAAAAA")
```
rkeene: We'll leave this debugging in for now as we ensure it's fixed, but maybe this should be a logged debug message in the future ?

a-urth:
@rkeene did changes according to all of your comments

PR #303, opened 2019-10-24
Add num_missing_tx, and have MarkMissing be an atomic CAS operation.

iwasaki-kenta:
api, actions, wctl: add num_missing_tx to ledger status response and cli
ledger, transactions: have MarkMissing be an atomic operation that returns true if the transaction has been successfully marked missing
transactions: remove BatchMarkMissing as we do not use it

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -59,7 +59,7 @@ func (t *Transactions) BatchAdd(block BlockID, transactions ...Transaction) {
 
 func (t *Transactions) add(block BlockID, tx Transaction) {
 	if t.height >= tx.Block+uint64(conf.GetPruningLimit()) {
-		fmt.Println(">>>>>>>>>>>> why??????")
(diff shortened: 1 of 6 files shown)


transactions.go
```diff
@@ -78,24 +78,23 @@ func (t *Transactions) add(block BlockID, tx Transaction) {
...
-func (t *Transactions) BatchMarkMissing(ids ...TransactionID) {
-	t.Lock()
-	defer t.Unlock()
+func (t *Transactions) markMissing(id TransactionID) bool {
+	_, exists := t.buffer[id]
 
-	for _, id := range ids {
-		t.markMissing(id)
+	if exists {
+		return false
```
a-urth: why return bool if its never used?

iwasaki-kenta: It's used in ledger.go to check that the transaction was atomically marked missing.

PR #305, opened 2019-10-24
Ensure that callbacks are registered before listening

rkeene:
Currently the callbacks for peers joining/leaving are set after the node starts listening for peers.  This creates a data race where `connLoop()` in Noise attempts to call those callbacks while the Wavelet node is registering them.

There is a separate PR ( https://github.com/perlin-network/noise/pull/254 ) to ensure that this does not cause a race condition, but it may be a good idea to ensure that these callbacks are registered before listening for connections.

### DIFF
--- a/api/mod.go
+++ b/api/mod.go
@@ -197,6 +197,8 @@ func (g *Gateway) StartHTTP(port int, c *skademlia.Client, l *wavelet.Ledger,
 
 	logger.Info().Int("port", port).Msg("Started HTTP API server.")
 
+	registerPeerCallbacks(c)
+
 	go g.start(ln, nil, c, l, k, kv)
 }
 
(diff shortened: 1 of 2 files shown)


iwasaki-kenta:
In this case, should we just leave a comment stating that setting callbacks should not be done in realtime in Noise? We can leave the mutex locking/unlocking up to the developer user in Noise then.

rkeene:
I also submitted a PR ( https://github.com/perlin-network/noise/pulls/254 ) to make it safe to do this in Noise, but in Wavelet's case we do want those callbacks set prior to accepting clients, otherwise the logged output may be confusing.

iwasaki-kenta:
> I also submitted a PR ( [perlin-network/noise#254](https://github.com/perlin-network/noise/pull/254) ) to make it safe to do this in Noise, but in Wavelet's case we do want those callbacks set prior to accepting clients, otherwise the logged output may be confusing.

Right, in that case we should just delegate any need for mutices or locking to the developers using Noise, and leave a warning that callbacks should always be set beforehand.

I think it's better that way - callbacks as an example shouldn't be expected to be changeable on-the-fly.

PR #307, opened 2019-10-25
Reduce mutex overhead in tight loop

hasyimibhar:
Profile trace:

Mutex trace:

With this change, `queried_bps` seems to be stable around 250 instead of steadily dropping.

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -2,10 +2,11 @@ package wavelet
 
 import (
 	"fmt"
-	"github.com/google/btree"
-	"github.com/perlin-network/wavelet/conf"
 	"math/big"
 	"sync"
+
+	"github.com/google/btree"
+	"github.com/perlin-network/wavelet/conf"
 )
 
 var _ btree.Item = (*mempoolItem)(nil)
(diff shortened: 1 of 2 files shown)


transactions.go
```diff
@@ -79,9 +80,6 @@ func (t *Transactions) add(block BlockID, tx Transaction) {
 // MarkMissing marks that the node was expected to have archived a transaction with a specified id, but
 // does not have it archived and so needs to have said transaction pulled from the nodes peers.
 func (t *Transactions) MarkMissing(id TransactionID) bool {
```
a-urth: i think it was convenient to have locking inside public function.
maybe leave this one as it is, but use private method with no locking in querying?

hasyimibhar: Yeah not having lock inside this function while others have is not consistent behavior. I think it's better to just not have locks at all within `Transactions`, and let the caller decide whether to lock or not.

hasyimibhar: Good idea, I'll go with `MarkMissingBulk` or `BatchMarkMissing`, since we already have `BatchAdd`.

hasyimibhar: Oh wait it's not as simple as that, since the for loop has to break once a single `MarkMissing` returns true.

a-urth: Thats true, still i think this logic should be within `transactions`. Then we can think of better name, like `FindMissing` or `TestForMissing` something like that.

hasyimibhar:
Closing this because it's on top of `snowball-update`, which is WIP. #309 is on top of `himitsu`, which makes it easier to merge.

PR #308, opened 2019-10-25
Fix concurrency bug in querying if request have failed.

a-urth:
There were 2 returns with no writes to `voteChan`, but we expected to have fixed number of values in that channel. So in case number was less than expected query process would stuck indefinitely waiting to read from channel.

+ simplify concurrency

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -738,99 +738,87 @@ func (l *Ledger) finalize(block Block) {
 func (l *Ledger) query() {
 	snowballK := conf.GetSnowballK()
 
-	if len(l.client.ClosestPeerIDs()) < snowballK {
-		// TODO not enough peers, what do to ?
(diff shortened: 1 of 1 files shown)


ledger.go
```diff
@@ -738,100 +738,88 @@ func (l *Ledger) finalize(block Block) {
...
 
 	req := &QueryRequest{BlockIndex: current.Index + 1}
 
-	for i := 0; i < cap(workerChan); i++ {
-		go func() {
-			for conn := range workerChan {
-				f := func() {
-					client := NewWaveletClient(conn)
+	for _, p := range peers {
+		wg.Add(1)
```
iwasaki-kenta: wg.Add(len(peers))

a-urth: there no difference, but i just prefer to keep `Add` and `Done` close to each other

hasyimibhar:
Reviewing.

PR #311, opened 2019-10-28
Shutdown investigation results.

a-urth:
- simplify concurrency for querying
 - add channel read timeout for voting results
 - change shutdown for stall detector
 - minor style fixes

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -810,12 +810,7 @@ func (l *Ledger) query() {
 	voters := make(map[AccountID]struct{}, len(peers))
 
 	for i := 0; i < cap(votes); i++ {
-		var vote *finalizationVote
-		select {
-		case vote = <-voteChan:
-		case <-time.After(conf.GetQueryTimeout()):
-			continue
-		}
+		vote := <-voteChan
 
 		if vote.voter == nil {
 			continue
--- a/ledger.go
+++ b/ledger.go
@@ -522,12 +522,9 @@ func (l *Ledger) PullMissingTransactions() {
 			req.TransactionIds = append(req.TransactionIds, txID[:])
 		}
 
-		var wg sync.WaitGroup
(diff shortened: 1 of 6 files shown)


ledger.go
```diff
@@ -809,13 +806,16 @@ func (l *Ledger) query() {
...
-	voters := make(map[AccountID]struct{}, snowballK)
+	votes := make([]*finalizationVote, 0, len(peers))
+	voters := make(map[AccountID]struct{}, len(peers))
 
 	for i := 0; i < cap(votes); i++ {
-		vote := <-voteChan
+		var vote *finalizationVote
+		select {
+		case vote = <-voteChan:
+		case <-time.After(conf.GetQueryTimeout()):
```
a-urth: i'd like this thing to be here at least for a while, because i reproduce it couple of times.
i have a feeling that since query function is running using metrics, metrics after being closed is not running the function and locks (either that was a race condition over metrics lock). nonetheless there was less than required results in vote channel.

rkeene:
I'm reviewing this now

ledger.go
a-urth: actually, thinking about that now, i think reason was in something else, maybe in having more peers than snowballK which resulted in channel being blocked on write inside goroutine.
currently i cannot reproduce it.

ledger.go
```diff
@@ -829,6 +824,8 @@ func (l *Ledger) query() {
 		votes = append(votes, vote)
 	}
 
+	wg.Wait()
```
rkeene: Can this actually ever wait for anything ?  It seems like the only thing it could wait on is for all the goroutines to finish, which the last thing they do is write to the channel which we consumed just prior to this

a-urth: actually you're right. since we read for certain number of values from channel we don't need wait group here at all

PR #312, opened 2019-10-28
Change block id to consist of transaction ids.

a-urth:
Since previously block id contained merkle root it changed depending on collapse result. Which led to double collapse call and wrong merkle for smart contracts.

### DIFF
--- a/block.go
+++ b/block.go
@@ -22,11 +22,7 @@ type Block struct {
 func NewBlock(index uint64, merkle MerkleNodeID, ids ...TransactionID) Block {
 	b := Block{Index: index, Merkle: merkle, Transactions: ids}
 
-	buf := bytes.NewBuffer(nil)
-	for _, id := range ids {
-		buf.Write(id[:])
-	}
-	b.ID = blake2b.Sum256(buf.Bytes())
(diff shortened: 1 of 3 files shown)


block.go
```diff
@@ -21,7 +21,12 @@ type Block struct {
 
 func NewBlock(index uint64, merkle MerkleNodeID, ids ...TransactionID) Block {
 	b := Block{Index: index, Merkle: merkle, Transactions: ids}
-	b.ID = blake2b.Sum256(b.Marshal())
+
+	buf := bytes.NewBuffer(nil)
```
iwasaki-kenta: We can set the capacity to be len(ids) * SizeTransactionID.

iwasaki-kenta:
Rather than modifying block IDs to be the hash of all transaction IDs by the way, it would be better to have the hash of all transaction IDs to be the key in which the `collapseTransactions` cache caches by.

We need to make sure that the merkle root of a block is 'fingerprinted' by a block ID, otherwise people can just change the contents of the merkle root of a block.

a-urth:
@iwasaki-kenta changed

PR #313, opened 2019-10-29
Fix Transaction Tag

AhmadMuzakkir:
The Transaction Tag value in the Wavelet's Himitsu is different than in the Smart Contract SDK because of the removal of Nop tag. The Smart Contract SDK still contains Nop tag.
Because of the Transaction Tag value is an incrementing number, this is causing it to have different meaning. 

All the smart contracts need to be recompiled with the updated Smart Contract SDK.

This PR updates the wasm by recompiling the wasm with the updated Smart Contract SDK.

PR to update the Smart Contract SDK:
https://github.com/perlin-network/smart-contract-rs/pull/9


AhmadMuzakkir:
I've reversed the new wasms since the Transaction Tag values remains same as before.

PR #315, opened 2019-10-31
Fix race conditions

hasyimibhar:
Detected in PR #314 while running tests. Originally part of that PR, but it's better to have a separate PR for fixing it.


hasyimibhar:
Somehow this test fails during the latest CI run:

PR #316, opened 2019-10-31
Refactor Vote implementation

AhmadMuzakkir:
Currently, there are a lot of pointers in the finalization and syncing votes calculation.
Since pointers will be allocated to the heap, this can contribute to the GC

Also, we use an interface to represent Vote. This makes the code a bit complex than it needs to be.

This PR refactors the vote calculation to convert the interface `Vote` to be struct and remove the pointers from the votes calculation. By doing so, this also simplifies the code.

This PR also adds more unit tests to increase the coverage of `snowball.go` and `vote.go` to 100%.

### DIFF
--- a/ledger.go
+++ b/ledger.go
@@ -625,9 +625,9 @@ func (l *Ledger) FinalizeBlocks() {
 					Int("num_transactions", len(proposedBlock.Transactions)).
 					Msg("Proposing block...")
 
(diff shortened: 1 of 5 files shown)


AhmadMuzakkir:
Closed, for now. 

A separate PR will be created for the unit tests.

PR #319, opened 2019-11-05
wctl refactoring: Concrete data structures and code cleanup

diamondburned:
Changes:
- Changes logging from key-value to concrete struct with proper unmarshaling

Known breaking changes:
- `num_accounts_in_store` is removed, as `num_accounts` has the same value.

Other minor changes:
- ~~Websocket ignores all events below `info`~~
- Logging methods are now very different (refer to `ledger.go` changes).
- API package is refactored.
- Events logged into the CLI are now scattered in `wavelet` and `api` packages. No more manually updating fields in `wctl`, `cli` and elsewhere.

This PR solves https://github.com/perlin-network/wavelet/pull/422.

~~! This PR is untested !~~
The PR now passes tests.


a-urth:
@diamondburned was this branch initially created from `himitsu`? because it seems like it contains a lot of commits from it.

diamondburned:
> @diamondburned was this branch initially created from `himitsu`? because it seems like it contains a lot of commits from it.

Yeah, it didn't make it in time for Himitsu. I'll rebase it to master soon.

accounts_events.go
```diff
@@ -0,0 +1,113 @@
...
+	return unmarshalAccountIDandUint64(v, a.AccountID, "reward", &a.Reward)
+}
+
+func marshalAccountIDandUint64(ev *zerolog.Event, id AccountID, ukey string, u uint64, updated string) {
+	ev.Hex("account_id", id[:])
+	ev.Uint64(ukey, u)
+	ev.Msg("Updated account " + updated + ".")
+}
+
+// yes, I know this is a terrible function name, it can't be helped
```
a-urth, line 105: i don't think this comment is necessary.

accounts_events.go
```diff
@@ -0,0 +1,113 @@
...
+func unmarshalAccountIDandUint64(v *fastjson.Value, id AccountID, ukey string, u *uint64) error {
```
a-urth, line 106: it definitely should be renamed at least to something containing `amount` instead of `uint64`.
also maybe 2 separate functions might be used - one for account and another one for amount? even though there will be two function calls, intention and naming will be much cleaner.

Issue #221, opened 2019-09-12
Build fails on 32 bit architecture


jakerr:
Build fails on 32 bit raspberry pi from source archive `wavelet-0.1.1`

The relevant error is:

```
...
# github.com/perlin-network/wavelet/avl
avl/node.go:371:17: constant 4294967295 overflows int
package main
avl/node.go:379:19: constant 4294967295 overflows int
avl/node.go:519:16: constant 4294967295 overflows int
avl/node.go:529:19: constant 4294967295 overflows int
```

The problem is that `len` returns a platform independent `int`.

Removing the checks:
```go
if len(n.key) > math.MaxUint32 {
    panic("avl: key is too long")
}
```
lets the build to succeed, though I imagine you'll want to find a platform independent way to keep those assertions.

rkeene:
This has been fixed as part of the work in #321 which has been merged to master and will be included in the next release.  Additionally, a 32-bit build is tested during CI.

PR #321, opened 2019-11-06
Minor fixes to make wavelet work on 32 bit arm

hasyimibhar:
- Fix compile errors due to math constants implicitly converted to int: https://github.com/golang/go/issues/24523#issuecomment-375973171
- Fix 64 bit atomic operations not working due to alignment issues: https://go101.org/article/memory-layout.html


hasyimibhar:
This PR should fix #221.

wctl/counter.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+	"unsafe"
+)
+
+type Counter struct {
+	n    uint64
+	lock sync.RWMutex
```
iwasaki-kenta: I feel this is a good time for us to introduce https://github.com/uber-go/atomic. Thoughts?

hasyimibhar: Good idea.

hasyimibhar:
Dont merge it yet. Seems like theres still an issue when using uber's atomic package.

hasyimibhar:
Okay it works now.

PR #323, opened 2019-11-07
Add abstract worker pool.

a-urth:
- use worker pool in query
- change config values for syncing


ledger.go
```diff
@@ -802,7 +810,9 @@ func (l *Ledger) query() {
 				vote.block = &block
 			}
 			l.metrics.queryLatency.Time(f)
-		}(p)
+		}
+
+		l.queueWorkerPool.Queue(func() { f(p) })
```
hasyimibhar: CI caught a data race somewhere here.

a-urth: I assume parameter enclosure for embeded function works differently

a-urth:
@hasyimibhar no race conditions now. test failed due to timeout though. locally i can't reproduce it, assumptions that it might because of worse performance on CI.

rkeene:
@a-urth Can you rebase this ?

a-urth:
@rkeene sure, i will

rkeene:
I'm reviewing this now

worker_pool.go
```diff
@@ -0,0 +1,48 @@
...
+				case <-wp.stop:
+					return
+				}
+			}
+		}()
+	}
+}
+
+func (wp *WorkerPool) Queue(job func()) {
+	wp.bus <- job
```
rkeene: Can we add a check here to do something if the WorkerPool has been closed ?   Maybe `panic` or something ?

worker_pool.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (wp *WorkerPool) Stop() {
```
rkeene: Should it be possible to call `wp.Start()` again after `wp.Stop()` has been called -- or should it just be an error to do so ?

a-urth: no restarts intended. worker pool lifecycle expected to be same as application using it.

worker_pool.go
a-urth: i think i should close `bus` channel and then here will be panic because of that

worker_pool.go
rkeene: Excellent, closing `wp.bus` should solve this as well

worker_pool.go
rkeene: Perfect

a-urth:
this one messed up, i'll recreate it

PR #324, opened 2019-11-08
Optimize query by caching the block

AhmadMuzakkir:
Currently, the query protocol always returns the entire Block data.
This is inefficient, in case of when a node has already received the same Block in the previous query. The inefficiency is also amplified by Snowball, as Snowball requires successive queries for the same Block.

This PR introduces a caching mechanism for the query protocol.
- Add a `Cache Block ID` field in the query request. The field is optional, where a node can send its cached Block ID for the peer. 
- Add a `Cache Valid` bool field in the query response. It would be true only when the `cache Block ID` is not nil and equal the to peer's proposed Block ID. In this case, the query response will not contain the block data.

It's up to the node whether to support caching. If a node does not want to cache, it can simply not send `cache Block ID`.

This PR also updates benchmark to set the number of workers to half of the number of logical CPU with a minimum of 2.
If your machine has 8 logical CPU, the number of workers would be 4 instead of 8.


utils.go
```diff
@@ -49,3 +50,48 @@ func SelectPeers(peers []*grpc.ClientConn, amount int) ([]*grpc.ClientConn, erro
...
+	for i := range ids {
+		if conn, err := c.Dial(ids[i].Address(), opts...); err == nil {
+			peers = append(peers, ClosestPeer{
+				conn: conn,
+				id:   ids[i],
+			})
+		}
+	}
+
+	activePeers := make([]ClosestPeer, 0, len(peers))
```
a-urth: now we kinda have this, pretty much identical, functionality in two places.

cmd/benchmark/flood.go
```diff
@@ -30,7 +30,10 @@ import (
 
 func floodTransactions() func(client *wctl.Client) ([]*wctl.TxResponse, error) {
 	return func(client *wctl.Client) ([]*wctl.TxResponse, error) {
-		numWorkers := runtime.NumCPU()
+		numWorkers := runtime.NumCPU() / 2
```
hasyimibhar: It doesn't feel right to simply divide it by 2, as it will return the wrong number of CPU if the processor does not have hyperthreading enabled. I think it's better to have this as a parameter, so that the user can choose how many workers to run for the benchmark (by default it will be `runtime.NumCPU`).

PR #330, opened 2019-11-12
Use cuckoo filter instead of bloomfilter.

a-urth:
Its faster than previous implementation using bloom filter, it should be more space efficient although marshaled structure is bigger, thats why number of elements is decreased.


AhmadMuzakkir:
My testing shows that, TPS is higher by using cuckoo filter than bloom filter.

I ran benchmark with 4 workers, the TPS was around ~5000.
With bloom filter, the TPS was around ~4000.

a-urth:
@rkeene @iwasaki-kenta could you please review it guys, when possible?

rkeene:
I'm reviewing this now

PR #334, opened 2019-11-18
Fix cli cleanup funcs slice being not modified.

a-urth:
While investigating problem with connect-disconnect test being stuck, I found that cleanup functions slice is passed as value and then modified by append, which result in new slice creation and none of the cleanup functions actually being applied.

Not the best fix, I personally don't like slices being appended like that, but its easiest one with no changes to current approach.

### DIFF
--- a/cmd/wavelet/events.go
+++ b/cmd/wavelet/events.go
@@ -8,15 +8,15 @@ import (
 )
 
 // converts a normal (func to close, error) to only an error
-func addToCloser(toClose []func()) func(f func(), err error) error {
+func addToCloser(toClose *[]func()) func(f func(), err error) error {
 	return func(f func(), err error) error {
-		toClose = append(toClose, f)
+		*toClose = append(*toClose, f)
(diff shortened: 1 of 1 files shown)


cmd/wavelet/events.go
```diff
@@ -8,15 +8,15 @@ import (
 )
 
 // converts a normal (func to close, error) to only an error
-func addToCloser(toClose []func()) func(f func(), err error) error {
+func addToCloser(toClose *[]func()) func(f func(), err error) error {
 	return func(f func(), err error) error {
-		toClose = append(toClose, f)
+		*toClose = append(*toClose, f)
```
AhmadMuzakkir: I think there's potential bug that can cause panic.

In the closure, we should check argument `f` is not nil before appending.

This is because if the Poll failed to Dial, it'll return (nil, err).

a-urth: nice catch. fixed

PR #337, opened 2019-11-18
Add linter

a-urth:
- add golangci-linter
 - fix all linting errors
 - fix some tests
 - skip connect-disconnect because its getting stuck indefinitely while running all the tests

NOTE:
 - its expected to have linter installed locally, there are 2 reasons for that: first - its not recommended to vendor it (and we do not vendor anything as of now), second - running it in docker messes up output, which is fine for CI but is not for running it locally
 - i'd expect to run it in docker in CI
 - connect-disconnect test is skipped since I think I've found problem, but i'll need to try one more thing and make fix in separate PR, since maybe its related to `noise`


a-urth:
I used locally linter 1.18, latest for mac from brew is 1.20.1 so i assume we'll use that one.
It has couple more linters so i'll need to add more changes to it.
This approach to actions works though

rkeene:
For the CI, can we have it convert known/accepted issues to Warnings ( https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions ) ?

a-urth:
@hasyimibhar not sure, since its step by step and if one fails, next one won't start
@rkeene sorry, what do you mean by known/accepted issues?

conf/conf.go
```diff
@@ -87,8 +87,7 @@ func defaultConfig() config {
 		blockTxLimit: 1 << 16,
 	}
 
-	switch sys.VersionMeta {
-	case "testnet":
+	if sys.VersionMeta == "testnet" {
```
rkeene: This is better as a `switch` because we'll soon be adding "mainnet" and "choasnet";  We can add them in there now just with the same changes for "chaosnet".

On the other hand, we should consolidate all the places that these parameters are changed so maybe we should just have a different PR for that and not delay this :-)

sys/const.go
```diff
@@ -65,186 +65,6 @@ var (
 
 	FaucetAddress = "0f569c84d434fb0ca682c733176f7c0c2d853fce04d95ae131d2f9b4124d93d8"
 
-	GasTable = map[string]uint64{
```
rkeene: Can we keep this with a `nolint` and open an issue to make it used ?

wctl/ws_accounts.go
```diff
@@ -113,22 +113,3 @@ func parseAccountStakeUpdated(c *Client, v *fastjson.Value) error {
 	}
 	return nil
 }
-
-func parseAccountRewardUpdated(c *Client, v *fastjson.Value) error {
```
rkeene: Should we figure out why this is no longer used ?

cmd/wavelet/update.go
```diff
@@ -218,35 +218,6 @@ func checkForUpdate(baseURL string, updateDirectory string, publicKeyPEM []byte,
 	return
 }
 
-func periodicUpdateRoutine(updateURL string) {
```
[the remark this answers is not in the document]
a-urth: ok

conf/conf.go
a-urth: as soon as more values will be added, switch will be required, so its fine

sys/const.go
a-urth: ok

wctl/ws_accounts.go
a-urth: do you know where it was used previously?

rkeene: I'm not sure -- it may have never been used.  We don't have to add it back but we should create an issue to track down whether or not it should be used.

AhmadMuzakkir: There's a mistake in PollAccounts that has gone undetected. 

We're suppose to use it to parse "reward_updated" event.

https://github.com/perlin-network/wavelet/blob/d27bb3167819201e17375c1dbfe7a9276c86a789/wctl/ws_accounts.go#L26-L27.

rkeene: Excellent find !

PR #332, opened 2019-11-14
Test all packages

hasyimibhar:
We have not been testing all packages (probably due to a typo in `workflow.yml`). This PR makes sure tests in all packages are run, and also fixes test errors that were not caught due to the tests not being run previously.

PR #346, opened 2019-11-19
Add abstract worker pool.

a-urth:
Recreation of #323


genesis.go
```diff
@@ -165,7 +165,7 @@ func restoreFromJSON(tree *avl.Tree, json []byte) error {
 	return nil
 }
 
-func restoreFromDir(tree *avl.Tree, dir string) error {
+func restoreFromDir(tree *avl.Tree, dir string) error { // nolint:gocognit,gocyclo
```
hasyimibhar: Is there a guideline when we should add `nolint`? In my PR, I extracted each of the `ext` switch statement case into [separate functions](https://github.com/perlin-network/wavelet/pull/332/files#diff-656e12979c589f523eba2358b8f69720) to remove the lint errors. But honestly I'm not sure if it's better, because each function has to take 4-5 variables, and most of the variables are modified by `restoreFromDir`, which I feel makes it more prone to errors (I prefer functions to be "pure").

rkeene: Right now we should just take linting errors as a strong hint that what we are doing may be wrong, but if you disagree with the hint or can't come up with a clearer way to write the code, then feel free to use "`nolint`".  Conversely, when you see "`nolint`" when reviewing a PR, it might be a good idea to look closer and try to contribute suggestions on how it could be more clearly written.

PR #349, opened 2019-11-20
BadgerDB: Handle directory path and enable GC

AhmadMuzakkir:
BadgerDB expects the directory to be already existed. If the directory does not exist, BadgerDB will return an error.

Also, BadgerDB expects on the client to perform Garbage Collection.

This PR adds a sane handling of the path.
1. If the path does not exist, it will create the directory.
2. If the path exists, but is a file, it will return an error.

Also, this PR implements a goroutine to call the GC every 1 minute.


store/badger.go
```diff
@@ -49,21 +53,45 @@ func (b *badgerWriteBatch) Destroy() {
...
 	db  *badger.DB
+
+	// Stops gc goroutine when db is closed.
+	closeCh chan struct{}
+	closeWg sync.WaitGroup
 }
 
-func NewBadger(dir string) (*badgerKV, error) { // nolint:golint
+func NewBadger(dir string) (*badgerKV, error) { // nolint:golint,gocritic
+	if fi, err := os.Stat(dir); err == nil {
```
rkeene: The linter really wants this to be a switch statement but I think it makes more sense as written than as switch.  Is there an even more clear way to write this ?

One option might be to unconditionally call os.Mkdir() and ignore the return value, then let `badger.Open()` fail if the conditions are still wrong (as long as it returns a meaningful error message)

store/badger.go
```diff
@@ -132,6 +154,39 @@ func (b *badgerKV) CommitWriteBatch(batch WriteBatch) error {
...
+
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-b.closeCh:
+				return
+			case <-ticker.C:
+				for {
```
a-urth: why there is another for loop? can't we just make gc interval shorter?

store/badger.go
```diff
@@ -49,21 +54,38 @@ func (b *badgerWriteBatch) Destroy() {
...
 	db  *badger.DB
+
+	// Stops gc goroutine when db is closed.
+	closeCh chan struct{}
+	closeWg sync.WaitGroup
 }
 
 func NewBadger(dir string) (*badgerKV, error) { // nolint:golint
-	db, err := badger.Open(badger.DefaultOptions(dir).WithLogger(nullLog{}))
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
hasyimibhar: We should add a test to check that the directory is created, since this is now the expected behavior.

store/badger.go
AhmadMuzakkir: It's an optimization in the documentation.
https://github.com/dgraph-io/badger#garbage-collection

PR #353, opened 2019-11-21
Badger: enable GC and upgrade to v2

AhmadMuzakkir:
This PR enables BadgerDB GC with an interval of 1 minute.

Also, upgrade to v2. Seems like there are bug fixes in v2.
https://github.com/dgraph-io/badger/blob/master/CHANGELOG.md


hasyimibhar:
The compile error on i386 seems like it is already fixed on master: https://github.com/dgraph-io/badger/pull/1118

hasyimibhar:
@AhmadMuzakkir I updated `go.mod` to temporarily use [our fork of Badger](https://github.com/perlin-network/badger), which has the i386 build issue fix patch applied on `v2.0.0` tag. Once the fix has been added to a tagged release, we should revert back to using dgraph's repository.

store/badger.go
```diff
@@ -49,21 +54,39 @@ func (b *badgerWriteBatch) Destroy() {
...
+	b := &badgerKV{
+		dir:     dir,
+		db:      db,
+		closeCh: make(chan struct{}),
+	}
+
+	b.gc(1 * time.Minute)
+
+	return b, nil
+
```
rkeene: The linter hates this new line 😄

store/badger.go
```diff
@@ -132,6 +154,39 @@ func (b *badgerKV) CommitWriteBatch(batch WriteBatch) error {
...
+				for {
+					// Check for close signal, in case it's closed during GC loop.
+					select {
+					case <-b.closeCh:
+						return
+					default:
+					}
+
+					// When there's no more GC, an error will be returned.
+					err := b.db.RunValueLogGC(0.5)
```
jarifibrahim: I would suggest you reduce the `discardRatio`. A ratio of `0.5` means that `Delete the file only if 50% of the file contains invalid data`. It would be difficult to get a discard ration of 0.5 unless you have equal number of read/write happening.

store/badger.go
```diff
@@ -132,6 +154,39 @@ func (b *badgerKV) CommitWriteBatch(batch WriteBatch) error {
...
+					if err != nil {
```
jarifibrahim: The `RunValueLogGC` might return an error `ErrNoRewrite` if it didn't perform any rewrite. In this case, you might want to retry the `GC`.

AhmadMuzakkir: Thank you for your help.

I'm sorry, I'm afraid I don't understand.

My understanding is that if the error is nil, we should retry. Otherwise, don't retry.

jarifibrahim: When `RunValueLogGC(..)` is called, it can return
1. `nil` which means GC was successful and a `vlog` file was removed.
2. `ErrNoRewrite` which means it found a `vlog` file to GC but it couldn't GC it because it had less amount of stale data (stale data less than discard ratio)
3. Any other error which occurred while performing garbage collection.

Case 1 is successful GC, case 2 is retry GC and case 3 is something went wrong with GC someone should look at it case.

AhmadMuzakkir: Got it.

In case of `ErrNoRewrite`, we will retry GC in the next cycle.

Currently, when the GC cycle (as for now, it's every 1 minute) is triggered, we would repeatedly call `RunValueLogGC(...)` until an error is returned. 
After which, we would wait until the next GC cycle to repeatedly call `RunValueLogGC(...)` again until an error is returned.

Below is how we handle the cases:
Case 1 (no error): Immediately call `RunValueLogGC(...)` again.
Case 2 (`ErrNoRewrite`): Wait for the next GC cycle to call `RunValueLogGC(...)` again..

store/badger.go
AhmadMuzakkir: Changed to 0.1.

PR #354, opened 2019-11-21
Merge coverage result from multiple test runs

hasyimibhar:
In PR #332, I separate tests into 2 runs: 1 for unit tests, and 1 for integration tests. The second test run currently overwrites the coverage report from the first test run. This PR merges the 2 coverage results instead.


a-urth:
@hasyimibhar  workflow file is malformed

PR #362, opened 2019-11-25
Optimize logging collapse results

AhmadMuzakkir:
There's a serious performance issue on the logging of collapse results. 
This performance issue can be severe when there are more than 10,000 transactions in a block. And it can have cascading impact on the performance of finalization (at least on my machine).

As you can see below, it can take more than 1 second to log a collapse result.
(The number in brackets is the number of transactions)

```
»»» logging collapse results (3946) took 76.556821ms
»»» logging collapse results (24215) took 472.709766ms
»»» logging collapse results (3927) took 69.653531ms
»»» logging collapse results (20420) took 470.434397ms
»»» logging collapse results (8291) took 177.229131ms
»»» logging collapse results (26042) took 503.408222ms
»»» logging collapse results (22277) took 455.032903ms
»»» logging collapse results (21466) took 492.797458ms
»»» logging collapse results (24661) took 551.213992ms
»»» logging collapse results (25779) took 657.309561ms
»»» logging collapse results (30573) took 715.361155ms
»»» logging collapse results (24427) took 603.267018ms
»»» logging collapse results (38552) took 869.41173ms
»»» logging collapse results (28388) took 2.282205659s
»»» logging collapse results (52269) took 2.0052882s
»»» logging collapse results (45526) took 688.535743ms
```

There are two issues with the logging that are causing performance issue.
1. The logging is a blocking process. Each transaction will be be serialized into json and write into the writers.

2. The logger will write on all the writers, even the writer's module does not match with the message's module. Currently, we have two writers.
I believe zerologger will filter the messages during the output.

However, I don't think there's any performance issue with the JSON serialization of zerolog library.

We did not notice this issue because the log was only recently enabled.

This PR fixes the performance issue by using buffer and write only to filtered writers.
We treat the `tx` event as a special case of logging, by not using zerolog.
The buffer is set 1024, this should give enough time for the writers to catch up, especially the `api` module (`websocket`) to handle the messages.

Below is the improved performance
```
»»» logging collapse results 5000 took 9.690248ms
»»» logging collapse results 27360 took 60.670627ms
»»» logging collapse results 4436 took 7.36232ms
»»» logging collapse results 22118 took 74.11094ms
»»» logging collapse results 4902 took 9.104645ms
»»» logging collapse results 21460 took 35.103805ms
»»» logging collapse results 1120 took 2.097708ms
»»» logging collapse results 13728 took 22.835278ms
»»» logging collapse results 9108 took 15.836429ms
»»» logging collapse results 5850 took 12.59921ms
»»» logging collapse results 8986 took 15.298364ms
»»» logging collapse results 7211 took 12.542619ms
»»» logging collapse results 6409 took 22.77037ms
»»» logging collapse results 6344 took 12.720952ms
»»» logging collapse results 9889 took 17.036898ms
»»» logging collapse results 3167 took 5.38616ms
```

### DIFF
--- a/log.go
+++ b/log.go
@@ -20,22 +20,143 @@
 package wavelet
 
 import (
+	"encoding/hex"
+	"sync"
+	"time"
+
+	"github.com/perlin-network/wavelet/conf"
 	"github.com/perlin-network/wavelet/log"
+	"github.com/valyala/fastjson"
 )
 
-func logEventTX(event string, tx *Transaction, other ...interface{}) {
-	logger := log.TX(event)
-	log := logger.Log().
-		Hex("tx_id", tx.ID[:]).
-		Hex("sender_id", tx.Sender[:]).
-		Uint8("tag", byte(tx.Tag))
+// CollapseResultsLogger is used to write CollapseResults to the logger's writers.
+//
+// It writes directly into the writers without going through zerolog.
+// The reason is that, zerolog will write into all the writers, even the writer's module does not match with the message's module.
+//
+// It also has a buffer to prevent blocking, as writing all the transactions in a collapse result may take sometime.
+// A collapse result may contain tens of thousands of transactions.
+type CollapseResultsLogger struct {
+	arena         *fastjson.Arena
+	mod           []byte // "tx"
+	eventApplied  []byte // "applied"
+	eventRejected []byte // "rejected"
+	timeLayout    string // "2006-01-02T15:04:05Z07:00"
 
-	for _, o := range other {
-		switch o := o.(type) { // nolint:gocritic
-		case error:
-			log = log.Err(o)
-		}
+	bufTxID     []byte
+	bufSenderID []byte
+	bufTime     []byte
+
+	// Buffer for a batch of messages (many transactions)
+	bufBatch [][]byte
+
+	flushCh chan [][]byte
+
+	stopWg sync.WaitGroup
+	stop   chan struct{}
+	closed bool
+}
+
+func NewCollapseResultsLogger() *CollapseResultsLogger {
+	c := &CollapseResultsLogger{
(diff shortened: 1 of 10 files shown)


log.go
```diff
@@ -20,22 +20,143 @@
...
+		mod:           []byte("tx"),
+		eventApplied:  []byte("applied"),
+		eventRejected: []byte("rejected"),
+		timeLayout:    "2006-01-02T15:04:05Z07:00",
+		bufTxID:       make([]byte, hex.EncodedLen(SizeTransactionID)),
+		bufSenderID:   make([]byte, hex.EncodedLen(SizeAccountID)),
+		bufTime:       make([]byte, 0, 64),
+
+		bufBatch: make([][]byte, 0, conf.GetBlockTXLimit()/4),
+		flushCh:  make(chan [][]byte, 1024),
```
a-urth: https://github.com/uber-go/guide/blob/master/style.md#channel-size-is-one-or-none

log_test.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+	txRejected := NewTransaction(keys, 0, 0, sys.TagTransfer, payload)
+	results.rejectedCount++
+	results.rejected = append(results.rejected, &txRejected)
+	results.rejectedErrors = append(results.rejectedErrors, errors.New("error rejected"))
+
+	logCh := make(chan []byte, 2)
+	log.SetWriter("tx_write_test", writerFunc(func(p []byte) (n int, err error) {
+		logCh <- p
```
hasyimibhar: The reason the test gets stuck is because the SetWriter persists between tests. Because of that, tests that run after this test will write to logCh, which causes it to block because no one is reading from the channel anymore.

The fix is to just clear the log writer at the end of this test.

PR #380, opened 2019-11-30
Remove account nonce.

a-urth:
and use timestamp for transaction creation

#378


a-urth:
@rkeene merge it please

PR #385, opened 2019-12-03
Fix Dockerfile

rkeene:
As part of adding BadgerDB support, CGO can be used for ZSTD compression;  Disable it while building the Docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ COPY go.sum /src/go.sum
 RUN (cd /src; go mod download)
 
 ADD . /src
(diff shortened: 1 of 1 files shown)


iwasaki-kenta:
I think we need to think this through a bit before merging it in: does this imply that databases which used zstd compression are incompatible with Wavelet nodes that are deployed on Docker containers?

rkeene:
Yes, that would be true.  This is also how we compile release binaries.

https://github.com/perlin-network/wavelet/blob/6d5c79ebb9c754417a975b4c908691a773cd80f1/scripts/helper.sh#L60-L61

Issue #382, opened 2019-12-02
After nonce removal, ensure newly started nodes can identify replayed transactions


rkeene:
As a follow-on to PR #380, on startup, pull blocks from the DB and add their transaction IDs to list of things we do not accept to avoid replay attacks.

PR #387, opened 2019-12-03
Store transactions to db along with stored blocks.

a-urth:
#382


iwasaki-kenta:
Rather than attempting to download transactions from previous blocks stored, why not instead have `Blocks` upon saving a block store all transactions down to disk?

a-urth:
Initially that way - download transactions on start - was easier to implement since we already have both parts: transaction ids stored in db and mechanism to download missing transactions. Now, considering that we need to separate downloading of already finalized transactions that might be not the case. Also this approach won't help in case all nodes will restarted at the same time, which unlikely but still possible.
But then I want to ask why we never saved transactions to db before?
@iwasaki-kenta

testutil.go
```diff
@@ -281,6 +276,11 @@ func NewTestLedger(t testing.TB, cfg TestLedgerConfig) *TestLedger {
 
 	kv, cleanup := store.NewTestKV(t, "level", path, kvOpts...)
 	ledger := NewLedger(kv, client, WithoutGC())
+
+	if ledger == nil {
+		panic("Error creating new test ledger")
```
iwasaki-kenta: It would be better to use `t.Fatal`` in this case.

a-urth: actually this won't help either since `log.Fatal` in `NewLedger` exits and since logger in not initialized there is no message.
I guess we need to change how ledger is created and return error from `NewLedger` and deal with it on upper level.

blocks.go
```diff
@@ -104,3 +104,14 @@ func (b *Blocks) GetByIndex(ix uint64) (*Block, error) {
 
 	return block, nil
 }
+
+func (b *Blocks) GetAll() []*Block {
```
iwasaki-kenta: Care to rename to `Clone()` and use `copy(blocks, b.buffer)` instead?

transactions.go
```diff
@@ -74,6 +74,17 @@ func (t *Transactions) BatchAdd(block BlockID, transactions []Transaction, verif
 	}
 }
 
+// BatchBufferAdd adds transactions to buffer without adding them to index
+// used on node's start
+func (t *Transactions) BatchBufferAdd(txs []*Transaction) {
+	t.Lock()
+	defer t.Unlock()
+
```
iwasaki-kenta: Care to rename to `BatchUnsafeAdd`? (adds transactions to the index without any validation).

blocks.go
a-urth: It doesn't clone actual structure, just returns blocks store.
Not sure whats the point in copying buffer, those blocks aren't modified and won't be modified by anything else.

transactions.go
a-urth: Why unsafe though?

blocks.go
iwasaki-kenta: The two should be equivalent; `copy` would just copy the pointers rather than the actual `Block` value itself.

transactions.go
iwasaki-kenta: Since it adds transactions directly to the buffer without any validation (unlike `BatchAdd`).

blocks.go
a-urth: yeah, so copy does same as append. i can change if you insist, but i see no reasons why.
what about name?

iwasaki-kenta: For the name, I reckon `Clone` would be better as we are cloning the buffer itself (as a slice of pointers). `copy` is just a little more idiomatic in this context so that we know we're taking a copy of the buffer.

a-urth: I guess it makes sense then to update style guide with preference of `copy` over `append`

PR #389, opened 2019-12-04
Fix failing genesis tests

hasyimibhar:
The tests are flaky because it does not properly wait for all nodes to sync, and uses `WaitUntilConsensus` instead of `WaitUntilBlock`.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,11 +28,11 @@ jobs:
       - name: Run unit tests
         run: make unit_test
 
-      - name: Run integration tests
-        run: make integration_test
-
       - name: Run genesis tests
         run: go test -v -tags=integration -parallel 1 . -run TestDump
+
+      - name: Run integration tests
+        run: make integration_test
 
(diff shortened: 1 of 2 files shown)


.github/workflows/workflow.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Run unit tests
         run: make unit_test
 
+      - name: Run genesis tests
```
iwasaki-kenta: Any reason on running genesis tests separately from integration/unit tests?

hasyimibhar: No reason, I'm just experimenting with Github CI. For some reason, running it separately makes the whole thing pass. No idea why.

iwasaki-kenta: Does not running it separately cause the test to be flaky then?

hasyimibhar: If you run the tests of each package separately (i.e. `go test .`, `go test ./debounce`, etc.) instead of `go test ./...`, all of the tests will actually pass.

PR #390, opened 2019-12-04
Github CI improvements.

hasyimibhar:
Few improvements to speed up CI cycle:

- [x] Cache go installation (looks like this is already handled by the action)
- [x] Cache go modules
- [x] Cache linter installation (it takes 2 seconds to install the linter, so I don't think there's much room to improve)
- [x] Split into multiple jobs to take advantage of parallelization


hasyimibhar:
Before caching:


After caching:


Looks like the overhead of invoking the cache action nullifies the speed improvement gained from caching modules.

PR #392, opened 2019-12-04
[WIP] Use sharded map to improve concurrency

hasyimibhar:
Use sharded map in `Transactions` buffer to improve concurrency.

- [ ] Test correctness
- [ ] Try different sharding methods (e.g. big int)
- [ ] Design benchmark to show performance improvement

### DIFF
--- a/shard.go
+++ b/shard.go
@@ -4,25 +4,24 @@ import (
 	"sync"
 )
 
-// TransactionMap is a sharded map with TransactionID as its key
-type TransactionMap map[uint64]*transactionMapShard
-
-type transactionMapShard struct {
-	items map[TransactionID]interface{}
-	lock  *sync.RWMutex
-}
-
 const (
 	shardsCount = 64
 )
 
-func NewTransactionMap() TransactionMap {
-	shards := make(TransactionMap, shardsCount)
+// TransactionMap is a sharded map with TransactionID as its key
+type TransactionMap [shardsCount]*transactionMapShard
 
+type transactionMapShard struct {
+	sync.Mutex
+
+	items map[TransactionID]interface{}
+}
+
+func NewTransactionMap() TransactionMap {
+	var shards TransactionMap
 	for i := uint64(0); i < shardsCount; i++ {
 		shards[i] = &transactionMapShard{
 			items: make(map[TransactionID]interface{}, 2048),
-			lock:  new(sync.RWMutex),
 		}
 	}
 
@@ -30,40 +29,37 @@ func NewTransactionMap() TransactionMap {
 }
 
 func (c TransactionMap) Len() uint64 {
-	// Lock all shards
-	for i := uint64(0); i < shardsCount; i++ {
-		c[i].lock.RLock()
-		defer c[i].lock.RUnlock()
-	}
-
 	var l uint64
+
 	for i := uint64(0); i < shardsCount; i++ {
+		c[i].Lock()
 		l += uint64(len(c[i].items))
+		c[i].Unlock()
 	}
 
 	return l
 }
 
 func (c TransactionMap) Iterate(fn func(key TransactionID, value interface{}) bool) {
-	// Lock all shards
 	for i := uint64(0); i < shardsCount; i++ {
-		c[i].lock.RLock()
-		defer c[i].lock.RUnlock()
-	}
+		c[i].Lock()
 
-	for i := uint64(0); i < shardsCount; i++ {
 		for key, value := range c[i].items {
 			if !fn(key, value) {
+				c[i].Unlock()
 				return
 			}
 		}
+
+		c[i].Unlock()
 	}
 }
 
 func (c TransactionMap) Get(key TransactionID) (interface{}, bool) {
 	shard := c.GetShard(key)
-	shard.lock.RLock()
-	defer shard.lock.RUnlock()
+
+	shard.Lock()
+	defer shard.Unlock()
 
 	if tx, exists := shard.items[key]; !exists {
 		return nil, false
@@ -75,8 +71,8 @@ func (c TransactionMap) Get(key TransactionID) (interface{}, bool) {
 func (c TransactionMap) Put(key TransactionID, value interface{}) {
 	shard := c.GetShard(key)
 
-	shard.lock.Lock()
-	defer shard.lock.Unlock()
+	shard.Lock()
+	defer shard.Unlock()
 
 	shard.items[key] = value
 }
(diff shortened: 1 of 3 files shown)


shard.go
```diff
@@ -0,0 +1,97 @@
...
+		l += uint64(len(c[i].items))
+	}
+
+	return l
+}
+
+func (c TransactionMap) Iterate(fn func(key TransactionID, value interface{}) bool) {
+	// Lock all shards
+	for i := uint64(0); i < shardsCount; i++ {
+		c[i].lock.RLock()
```
iwasaki-kenta: It would be better to read-plock as we are going through each shard, rather than locking them simultaneously all at once.

shard.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"sync"
+)
+
+// TransactionMap is a sharded map with TransactionID as its key
+type TransactionMap map[uint64]*transactionMapShard
+
+type transactionMapShard struct {
+	items map[TransactionID]interface{}
+	lock  *sync.RWMutex
```
iwasaki-kenta: Wonder if we should use a pure `sync.Mutex` instead. For large number of readers `sync.RWMutex` would make sense, though the time it takes to lock/unlock for a `sync.RWMutex` in general is about 3 - 4x slower.

diamondburned: I don't see a need for a pointer to a mutex, just the mutex being in the struct is enough.

shard.go
```diff
@@ -0,0 +1,97 @@
...
+func (c TransactionMap) Len() uint64 {
+	// Lock all shards
+	for i := uint64(0); i < shardsCount; i++ {
+		c[i].lock.RLock()
+		defer c[i].lock.RUnlock()
+	}
+
+	var l uint64
+	for i := uint64(0); i < shardsCount; i++ {
+		l += uint64(len(c[i].items))
```
diamondburned: I think this could be locked, incremented, then unlocked afterwards right in one loop.

shard.go
```diff
@@ -0,0 +1,97 @@
...
+
+const (
+	shardsCount = 64
+)
+
+func NewTransactionMap() TransactionMap {
+	shards := make(TransactionMap, shardsCount)
+
+	for i := uint64(0); i < shardsCount; i++ {
+		shards[i] = &transactionMapShard{
```
diamondburned: Why not use an array?

hasyimibhar: My bad, didn't notice it's a map.

hasyimibhar:
So far, eyeballing using the `cmd/benchmark` tool shows that the performance is getting worse. While TPS is roughly the same, querying is way slower than before, and it looks like block finalization is getting slower and slower.

Before (highest TPS):
```
9:08AM INF Benchmarking...
accepted_tps: 2815.229572505504
downloaded_tps: 245.89731548175118
gossiped_tps: 0
queried_bps: 375.7994980237461
query_latency_max_ms: 2
query_latency_mean_ms: 0.19335618190661477
query_latency_min_ms: 0
received_tps: 245.89731438496207
```
Profile: [profile002.pdf](https://github.com/perlin-network/wavelet/files/3924657/profile002.pdf)

After (highest TPS):
```
9:14AM INF Benchmarking...
accepted_tps: 2857.280377661443
downloaded_tps: 0
gossiped_tps: 0
queried_bps: 199.67517094665232
query_latency_max_ms: 12
query_latency_mean_ms: 0.10695210826848248
query_latency_min_ms: 0
received_tps: 0
```
Profile: [profile001.pdf](https://github.com/perlin-network/wavelet/files/3924659/profile001.pdf)

iwasaki-kenta:
> So far, eyeballing using the `cmd/benchmark` tool shows that the performance is getting worse. While TPS is roughly the same, querying is way slower than before, and it looks like block finalization is getting slower and slower.
> 
> Before (highest TPS):
> 
> ```
> 9:08AM INF Benchmarking...
> accepted_tps: 2815.229572505504
> downloaded_tps: 245.89731548175118
> gossiped_tps: 0
> queried_bps: 375.7994980237461
> query_latency_max_ms: 2
> query_latency_mean_ms: 0.19335618190661477
> query_latency_min_ms: 0
> received_tps: 245.89731438496207
> ```
> 
> Profile: [profile002.pdf](https://github.com/perlin-network/wavelet/files/3924657/profile002.pdf)
> 
> After (highest TPS):
> 
> ```
> 9:14AM INF Benchmarking...
> accepted_tps: 2857.280377661443
> downloaded_tps: 0
> gossiped_tps: 0
> queried_bps: 199.67517094665232
> query_latency_max_ms: 12
> query_latency_mean_ms: 0.10695210826848248
> query_latency_min_ms: 0
> received_tps: 0
> ```
> 
> Profile: [profile001.pdf](https://github.com/perlin-network/wavelet/files/3924659/profile001.pdf)

Based on those benchmarks, what if you have `missing` use a normal map and leave `transactions` to using a sharded map?

hasyimibhar:
Benchmarks of different map (you can get results by running `go run benchmarks/shard.go`):

shard.go
```diff
@@ -0,0 +1,93 @@
...
+
+	shard.Lock()
+	defer shard.Unlock()
+
+	delete(shard.items, key)
+}
+
+func (c TransactionMap) GetShard(key TransactionID) (shard *transactionMapShard) {
+	// Use the first byte as the shard ID
+	shardID := uint64(key[0]) % shardsCount
```
iwasaki-kenta: Rather than use modulus, could you try it with this and re-run benchmarks?

`(uint64(key[0]) * uint64(shardsCount)) >> 32 `

iwasaki-kenta: Inspiration: https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/

This assumes that `shardsCount` is a power of 2 btw.

hasyimibhar: It's slower:

iwasaki-kenta: That seems ... strange; what about `uint64(key[0]) & uint64(shardsCount - 1)`? Also, since we can fit shards into a uint16, we can make the keys uint16 and do `uint16(key[0] & uint16(shardsCount - 1)`.

hasyimibhar: About the same as with modulo:

iwasaki-kenta: Okay - FYI, since you are just using key[0], you can limit to just using 256 shards.

hasyimibhar:
Turns out I got the benchmark wrong. In the current implementation in `master`, in `BatchMarkMissing`, we lock the map once, go through the list of ids, then unlock it. In the benchmark, I lock the map for each read in a tight loop, so it made it worse. After adjusting the benchmark so that it's locked outside of the loop, now it's showing that map with `sync.Mutex/sync.RWMutex` is still faster that my implementation of sharded map for our use case (reading multiple keys at once):

iwasaki-kenta:
In this case, let's close this PR for the time being and put it in the backlog. Let's investigate it once more after v0.3.

PR #396, opened 2019-12-05
Separate usage of the "api.host" flag

rkeene:
Use a separate flag to specify the host wctl should connect to versus… the one we want to allow for HTTPS API access


iwasaki-kenta:
Care to rename to `cli.host` and `cli.port` instead?

rkeene:
Renamed

iwasaki-kenta:
Thanks! LGTM.

PR #398, opened 2019-12-05
Pre-validate transactions based on current state.

AhmadMuzakkir:
This PR implements state pre-validations for transactions to make sure the transactions can be applied based on the current state.

The Send Transaction API handler will return the error if the pre-validation failed.

Some of the validations are:
1. Sender must have enough balance for any movement of money. e.g. to place stake, to transfer money, to spawn contracts, etc.
2. For contract transaction, check the recipient address is actually a contract address.

TODO
- [x] Unit tests


api/mod.go
```diff
@@ -354,9 +354,19 @@ func (g *Gateway) sendTransaction(ctx *fasthttp.RequestCtx) {
 
 	if !tx.VerifySignature() {
 		g.renderError(ctx, ErrBadRequest(errors.New("bad signature")))
+
+		return
+	}
+
+	snapshot := g.ledger.Snapshot()
+
+	if err := wavelet.ValidateTransaction(snapshot, tx); err != nil {
```
a-urth: why we validate tx here instead passing `validateState=true` to `Add` method?

ledger.go
```diff
@@ -635,10 +648,14 @@ func (l *Ledger) PullMissingTransactions() {
 				continue
 			}
 
+			if err := ValidateTransaction(snapshot, tx); err != nil {
```
a-urth: same here - why validate it here instead passing true as validation flag?

api/mod.go
iwasaki-kenta: I feel that we should always validate the state of transactions within `(*Ledger).AddTransaction`. Rather than pass an optional boolean flag `validateState`, we should enforce it by default.

transactions.go
```diff
@@ -58,9 +58,11 @@ func (t *Transactions) BatchAdd(block BlockID, transactions []Transaction, verif
 		filtered := transactions[:0]
 
 		for i := range transactions {
-			if transactions[i].VerifySignature() {
-				filtered = append(filtered, transactions[i])
+			if !transactions[i].VerifySignature() {
```
iwasaki-kenta: Hm, could signature verification be moved to PrevalidateTransaction?

AhmadMuzakkir: Done.

PR #399, opened 2019-12-05
Optimize cuckoo filter, and remove redundant config parameters.

iwasaki-kenta:
cuckoo, ledger, protocol, mod: optimize cuckoo filter to reduce number of rounds of hashing required, utilize jenkins hashing, and to improve marshal/unmarshaling speeds
conf, cmd/wavelet: remove unnecessary config parameters for bloom filter params


internal/cuckoo/filter.go
```diff
@@ -0,0 +1,169 @@
...
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math/rand"
+)
+
+const (
+	BucketSize           = 4
+	NumBuckets           = 524288
```
a-urth: why number of buckets is fixed? shouldn't it be dependent on number of store values?

internal/cuckoo/filter_test.go
```diff
@@ -0,0 +1,170 @@
+package cuckoo
```
a-urth: we use build tags to distinguish unit and integration tests

internal/cuckoo/filter.go
iwasaki-kenta: It's a fixed parameter actually - the number of buckets should be consistent across all nodes.

internal/cuckoo/filter_test.go
iwasaki-kenta: Yep, will fix.

Issue #394, opened 2019-12-04
Restore blocks tests (previous rounds)


a-urth:
Before merging `himitsu` branch, and changing all rounds to blocks, we had same functionality to store particular number of rounds in file called `rounds.go`. This file was renamed to `blocks` and modified correspondingly. Although tests for that functionality which were in the file `rounds_test.go` for some reason were deleted.
I think its good idea to restore them and modify same way as main functionality was changed.

PR #401, opened 2019-12-05
Add blocks tests.

a-urth:
#394

### DIFF
--- a/blocks_test.go
+++ b/blocks_test.go
@@ -1,0 +1,107 @@
+// +build unit
+
+package wavelet
+
+import (
+	"crypto/rand"
+	"github.com/perlin-network/wavelet/store"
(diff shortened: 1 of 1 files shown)


blocks_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"crypto/rand"
+	"github.com/perlin-network/wavelet/store"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestNewRounds(t *testing.T) {
```
iwasaki-kenta: Care to rename the prefixes to be Blocks instead of Rounds?

a-urth: sry, my bad

PR #402, opened 2019-12-06
Check block height of all transactions listed under a queried block, and cleanup vote methods and don't allow NewBlock() to return errors.

### LINKED ISSUE #381: Check block height of all transactions listed under a queried block (to see if they really are valid proposable transactions).
Ensure that if the block height for a transaction is very high that it's not included in a block until that block height is reached.

iwasaki-kenta:
Fixes #381.


ledger.go
```diff
@@ -937,42 +936,64 @@ func (l *Ledger) query() {
...
+		}
+
+		transactions, err := l.transactions.BatchFind(vote.block.Transactions)
+		if err != nil {
+			vote.block = nil
+			continue ValidateVotes
 		}
 
+		// Validate the height recorded on transactions inside the block proposal.
+		for _, tx := range transactions {
```
hasyimibhar: We should add a test that checks if this works (e.g. modify one of the node's latest block height, and then check to make sure the node's proposed block never gets accepted):

iwasaki-kenta: Gotcha, will work on it.

block.go
```diff
@@ -40,24 +34,18 @@ func (b *Block) GetID() string {
...
-		return nil, errors.Wrap(err, "error marshaling transactions num")
-	}
+	binary.BigEndian.PutUint64(buf[0:8], b.Index)
+	copy(buf[8:8+SizeMerkleNodeID], b.Merkle[:])
+	binary.BigEndian.PutUint32(buf[8+SizeMerkleNodeID:8+SizeMerkleNodeID+4], uint32(len(b.Transactions)))
 
-	for _, id := range b.Transactions {
-		buf.Write(id[:])
+	for i, id := range b.Transactions {
+		copy(buf[8+SizeMerkleNodeID+4+i*SizeTransactionID:8+SizeMerkleNodeID+4+i*SizeTransactionID+SizeTransactionID], id[:])
```
a-urth: maybe it makes sense to create separate variables for parts of slice indexes? like
```
prefix := 8+SizeMerkleNodeID+4
for i, id := range b.Transactions {
    copy(buf[prefix+i*SizeTransactionID:prefix+(i+1)*SizeTransactionID]
}
```

ledger.go
```diff
@@ -1667,31 +1622,31 @@ func (l *Ledger) resetTransactionsSyncIndex() {
 }
 
 func (l *Ledger) loadTransactions() error {
-	txNum := 0
+	count := 0
 
 	blocks := l.blocks.Clone()
 	for _, b := range blocks {
-		txs, err := LoadTransactions(l.db, b.Transactions)
+		transactions, err := LoadTransactions(l.db, b.Transactions)
```
a-urth: we already have `transactions` and this variable used in a pretty small blocks, thats why `txs` was used.
same for `txNum`, its much better represents what this counter means than just `count`

transactions.go
```diff
@@ -76,7 +76,7 @@ func (t *Transactions) BatchAdd(block BlockID, transactions []Transaction, verif
 
 // BatchUnsafeAdd adds transactions to buffer without adding them to index
 // used on node's start
-func (t *Transactions) BatchUnsafeAdd(txs []*Transaction) {
+func (t *Transactions) BatchUnsafeAdd(txs ...*Transaction) {
```
a-urth: whats the point if we are adding slice here anyways?
`BatchAdd` also works with slice btw

ledger.go
```diff
@@ -1714,3 +1669,64 @@ func (l *Ledger) storeTransactions(stored *Block, evicted *Block) error {
...
+		if err != nil {
+			vote.block = nil
+			continue ValidateVotes
+		}
+
+		// Validate the height recorded on transactions inside the block proposal.
+		for _, tx := range transactions {
+			if vote.block.Index >= tx.Block+uint64(conf.GetPruningLimit()) {
+				vote.block = nil
+				continue ValidateVotes
```
a-urth: thats the only place where you need to use continue to label, in all other places its one for loop and just continue can be used.
also, maybe instead of nested loop and label separate function with return can be used?

ledger.go
```diff
@@ -1714,3 +1669,64 @@ func (l *Ledger) storeTransactions(stored *Block, evicted *Block) error {
...
 }
+
+// filterInvalidVotes takes a slice of (*finalizationVote)'s and filters away
+// ones that are invalid with respect to the current nodes state.
+func (l *Ledger) filterInvalidVotes(current *Block, votes []Vote) {
+ValidateVotes:
+	for _, vote := range votes {
+		vote := vote.(*finalizationVote)
+
+		// Ignore nil block proposals.
```
a-urth: not sure about those comments, if statements already doing pretty good job explaining whats going on in each of them

block.go
AhmadMuzakkir: I've already made that changes.

transactions.go
iwasaki-kenta: It's just to allow for multiple methods of providing a slice of `*Transaction`'s to make tests a little cleaner. Hm, I believe `BatchAdd` in the past used to accept a variadic slice of `*Tranasction`'s, but upon adding in `verifySignature bool` we had to remove it.

ledger.go
iwasaki-kenta: Some of the if statements are a little confusing, so the comments are just to give some clarifications.

For example: `tx.Block >= current.Index + uint64(conf.GetPruningLimit())`.

transactions.go
a-urth: Yes, I've changed since it was used only with slice. Same here. And even name `Batch` implies that were will be multiple values sent, so slice seems reasonable.

ledger.go
iwasaki-kenta: I think the context for the variable names is already clear in the function name `loadTransactions` which is why I simplified the variable names.

transactions.go
iwasaki-kenta: Okay, will revert - it makes sense because these batch operations are important to distinguish from single operations (where we lock once vs many times).

ledger.go
iwasaki-kenta: Hm, I think it would be a bit odd to separate `filterInvalidVotes` into two separate functions where one of the functions gets called on each vote and nullifies `vote.block`. Hence why I kept it to just one single function and labeled the loop instead.

PR #404, opened 2019-12-06
Re-enable periodic update routine

### LINKED ISSUE #343: Re-add call to periodicUpdateRoutine

rkeene:
This resolves #343


iwasaki-kenta:
I don't think we should enable it yet - the auto-update infrastructure (either CI/server) still needs to be set up.

PR #406, opened 2019-12-06
Cleanup modules, simplify state collapse caching, and make certain marshal methods zero-alloc.

iwasaki-kenta:


avl/lru.node.go
```diff
@@ -29,6 +28,12 @@ import (
 	"sync"
 )
 
+//nolint:lll
```
a-urth: whats that?

block.go
```diff
@@ -50,11 +51,14 @@ func (b Block) Marshal() []byte {
 
 	n += 4
 
-	for _, id := range b.Transactions {
-		copy(buf[n:n+SizeTransactionID], id[:])
+	var ids []byte
 
-		n += SizeTransactionID
-	}
+	sh := (*reflect.SliceHeader)(unsafe.Pointer(&ids))
```
a-urth: is it worth it using such obscure piece of code for performance?
At least comment should be added, since now its really hard to understand why its done this particular way.

ledger.go
```diff
@@ -321,28 +318,6 @@ func (l *Ledger) Find(query string, max int) (results []string) {
 	return append(results, l.indexer.Find(query, count)...)
 }
 
-// PushSendQuota permits one token into this nodes send quota bucket every millisecond
-// such that the node may add one single transaction into its graph.
-func (l *Ledger) PushSendQuota() {
```
a-urth: aren't we planning to use it later?

ledger.go
```diff
@@ -184,12 +185,10 @@ func NewLedger(kv store.KV, client *skademlia.Client, opts ...Option) (*Ledger,
...
 
-		cacheCollapse: NewCollapseLRU(16),
-		fileBuffers:   filebuffer.NewPool(sys.SyncPooledFileSize, ""),
-
-		sendQuota: make(chan struct{}, 2000),
+		stateLRU: NewStateLRU(16),
+		filePool: filebuffer.NewPool(sys.SyncPooledFileSize, ""),
 
-		transactionsSyncIndex: cuckoo.NewFilter(),
+		filter: cuckoo.NewFilter(),
```
a-urth: previous naming was explaining what this structure is used for rather than just what this structure is, which can be determined by type btw.

lru.state.go
```diff
@@ -29,29 +28,35 @@ import (
 	"sync"
 )
 
-type CollapseLRU struct {
+//nolint:lll
```
a-urth: why there are 3 nolint directives?

ledger.go
iwasaki-kenta: It would be best for us to keep the rate limiting outside of `ledger` and inside the APIs instead.

avl/lru.node.go
iwasaki-kenta: I'm not 100% sure why the nolint tag was added - it might be because I'm using another genny installation (can switch to another one to double-check).

block.go
iwasaki-kenta: Yep, there is a noticeable performance boost as marshaling is done often and blocks often contain thousands of transaction IDs. I'll leave a comment though.

Issue #395, opened 2019-12-04
Fix transaction syncing after state sync.


a-urth:
There is problem with double spending after state syncing.
Imagine node which getting restarted without having database.
After syncing state it'll start syncing transactions and since it'll have current block but none of the transactions it'll sync all of the previously applied transactions.
Then it'll get them into block and start proposing it. Since other nodes already have those transaction in some circumstances they may vote in favor of that block and finalize it.
This will result into transactions previously applied being applied again.

I'm not sure that there is an easy solution aside using db always (considering #382 is done).

iwasaki-kenta:
After state syncing, we should have received roughly what the majority of nodes believe is the latest block _B_.

None of the nodes should finalize block _B_ ever again because block _B_ either:
1) the nodes have a different block stored at block _B_'s height.
2) the nodes already have block _B_.
3) the block is at a height that should have already been pruned away.

A change to transaction syncing needs to be done. Now that we have all transactions stored in the database as well, what should happen is that upon state syncing, we also download transactions for block _B_.

a-urth:
Block B won't be finalized again, I was saying that it would be new block containing already applied transactions.
Downloading transactions just for block B (block received from syncing) won't help since there are transactions for at least `syncAfterBlock` and at most `pruningLimit`-1 blocks more, which will be synced and restarted node is not aware about whether they are new or already have been applied.
We can to not download transactions from block B by explicitly adding them to sync tx index, or download them without adding to proposal index, since we know that they are already in finalized block, but we need to do the same for previous `pruningLimit` number of blocks as well.
Alternatively we need to change how we sync transactions - sync separately ones in finalized blocks and ones which are "new". But nonetheless we'll need to sync previous blocks before block B as well.

iwasaki-kenta:
> Block B won't be finalized again, I was saying that it would be new block containing already applied transactions.
> Downloading transactions just for block B (block received from syncing) won't help since there are transactions for at least `syncAfterBlock` and at most `pruningLimit`-1 blocks more, which will be synced and restarted node is not aware about whether they are new or already have been applied.
> We can to not download transactions from block B by explicitly adding them to sync tx index, or download them without adding to proposal index, since we know that they are already in finalized block, but we need to do the same for previous `pruningLimit` number of blocks as well.
> Alternatively we need to change how we sync transactions - sync separately ones in finalized blocks and ones which are "new". But nonetheless we'll need to sync previous blocks before block B as well.

[...]

PR #411, opened 2019-12-09
Fix state syncing.

a-urth:
#395 #382


transactions.go
```diff
@@ -138,10 +153,8 @@ func (t *Transactions) ReshufflePending(next Block) []TransactionID {
 
 	// Delete mempool entries for transactions in the finalized block.
 
-	lookup := make(map[TransactionID]struct{})
-
 	for _, id := range next.Transactions {
-		lookup[id] = struct{}{}
+		t.finalized[id] = struct{}{}
```
iwasaki-kenta: Rather than keep a separate `finalized` index, why not just check `t.buffer`? If I recall, `t.buffer` is persisted and loaded on startup with all finalized transactions from the database in `(*Ledger).loadTransactions()`.

protocol.go
```diff
@@ -287,3 +287,12 @@ func (p *Protocol) PullTransactions(
 
 	return res, nil
 }
+
+func (p *Protocol) GetBlock(ctx context.Context, req *GetBlockRequest) (*GetBlockResponse, error) {
```
iwasaki-kenta: Rather than add a separate `GetBlockRequest/GetBlockResponse`, care to use just `QueryBlockRequest/QueryBlockResponse`?

ledger.go
```diff
@@ -1019,426 +938,535 @@ func (l *Ledger) SyncToLatestBlock() { // nolint:gocyclo,gocognit
...
+			Msg("Failed to apply re-assembled diff to our ledger state. Restarting sync...")
 
-				cleanup()
+		return false
+	}
 
-				goto SYNC
-			}
+	// get all the blocks up until latest one
+	var oldestBlockIx uint64 = 1
```
iwasaki-kenta: Wasn't the objective to perform Snowball sampling and consensus over the last `PruningLimit` number of blocks, rather than just downloading all the latest blocks from `latestVoter`?

Downloading and trusting all the latest blocks from `latestVoter` is quite dangerous.

a-urth:
@hasyimibhar this shouldn't affect tps much, but i'll investigate. it might be that saving transactions to db affecting performance

protocol.go
a-urth: there is no `QueryBlockRequest` but there is `QueryRequest` and it contains field `cache_block_id` along with block index. same for response object, one used for querying has additional field in it.
i think its better to use different message objects here.

transactions.go
a-urth: `loadTransactions` loads transactions to buffer only in case they are stored in db. if node is restarted and then synced, those transactions won't be in db, but they will be synced. so we can mark them as finalized - since we only have id at this point - and then safely sync them along any other new transaction.
also in buffer we store all transactions, both finalized and not, so we need somehow to distinguish them.

protocol.go
iwasaki-kenta: Hm, I'd say it's best to just stick to 1 message object to make the RPC interface as minimal as possible. We can intentionally leave out cache_block_id here in this case.

transactions.go
iwasaki-kenta: Hm, okay :+1:.

ledger.go
a-urth: picking block just from the first voter is also dangerous.
i don't think its feasible to do it that way rn. we cannot send all blocks in a single request, so we need to send only block ids and sample those and then download blocks by ids from random peers.
i did it this way since we already relying on one voter by picking latest block by index, so it was logical to get other blocks from same voter.

protocol.go
a-urth: i don't think that reusing messages that are specific for some particular request is a good idea.
query messages already have extra information and specific name, if we'll need to add something more to them, then using those in `GetBlock` will be even more confusing.

iwasaki-kenta: QueryBlock is equivalent to asking for a block from a peer though - the only additional field is for caching.

ledger.go
iwasaki-kenta: Hm, rather than do it in a single request, I feel we should do it by:

1. If we are out of sync stop consensus.
2. Sample repetitively for the latest blocks contents from peer(s) using Snowball. Let's say we are now in-sync, knowing the block at height H.
3. Sample repetitively for height H+1, and do this for blocks at heights H+PruningLimit.
4. We now have PruningLimit number of blocks. Then continue on with consensus.

a-urth:
@hasyimibhar so i didn't notice any performance decrease for me.

ledger.go
a-urth: i don't understand why we should sample for PruningLimit number of blocks after latest one, unless its only to avoid getting already finalized transactions. And IMO its really big time to wait in order for single node to get back in sync.
Won't it be easier to sample for chain of blocks i.e. PruningLimit number of block ids like i suggested earlier?

iwasaki-kenta: Sampling for a chain of blocks is possible too, though we need to continually be sure to sample the next blocks that proceed to get finalized too in parallel.

If we sample the chain of the last PruningLimit blocks and the majority of nodes only keep the last PruningLimit blocks in their database, we may be unable to fetch the latestBlock.Height - PruningLimit block in time.

iwasaki-kenta:
Here's a simpler idea: store the last PruningLimit blocks transaction IDs in the state AVL tree. Consensus will finalize blocks, and we only need to sync to the latest block's state.

a-urth:
@iwasaki-kenta is going to implement proposed solution in separate PR branched out from this one.
@a-urth going to add to this PR check for pulling blocks after sync - if block is pruned on target node, its ok to skip it

hasyimibhar:
@a-urth My bad, there wasn't any performance decrease. Somehow when I benchmarked it that time, it was slow. I could not reproduce it anymore even if I checkout to the specific commit. I guess something else was running in the background when I benchmarked that day.

a-urth:
Closed in favor of #419
RIP my work on this and #387 prs...

PR #412, opened 2019-12-10
Keep peer block query responses in bounded LRU cache.

AhmadMuzakkir:
This PR will clear the `queryPeerBlockCache` when a block is finalized, since the cache is only valid for a block.


iwasaki-kenta:
Isn't the cache valid outside of a single block height? The security concern we need to address is what if the map gets too big - we need to "bound" the size of this cache.

Issue #421, opened 2019-12-12
WS events errors.


a-urth:
There are couple of unsupported ws events, which we either need to remove or make them supported.
`WS Error occurred. error: "Unsupported event: rejected"` happens on error during tx application which results into it being rejected.
And
`WS Error occurred. error: "Unsupported event: execute"` - not sure, but i think it happens on smart contract invocation.

a-urth:
Both mentioned events need to be supported.

PR #422, opened 2019-12-12
Temporary patch to 'Unsupported event'

diamondburned:
Unsupported events are now silenced. Fixes https://github.com/perlin-network/wavelet/issues/421.

When https://github.com/perlin-network/wavelet/pull/319 is merged, all events will be logged fully again.

### DIFF
--- a/log.go
+++ b/log.go
@@ -129,6 +129,7 @@ func (c *CollapseResultsLogger) Log(results *collapseResults) {
 func (c *CollapseResultsLogger) addTx(mod, event []byte,
 	timestamp time.Time, tag int,
 	txID []byte, sender []byte, logError error) {
+
 	o := c.arena.NewObject()
 
 	o.Set("mod", c.arena.NewStringBytes(mod))
--- a/wctl/wctl.go
+++ b/wctl/wctl.go
@@ -120,6 +120,7 @@ type Client struct {
 	// PollTransactions
 	OnTxApplied
 	OnTxGossipError
+	OnTxRejected
 	OnTxFailed
 
 	OnMetrics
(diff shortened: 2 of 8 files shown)


a-urth:
@diamondburned thats not exactly what i meant, sorry for misunderstanding.
So commenting handling unknown events is definitely not what we need.
I'm not sure whether we need those 2 events at all or they can be just log messages. First can be clarified with @iwasaki-kenta and @claudiucelfilip 

But how do you think much time it will take to add ws support (parsing mainly) for those two events? Thats in case we actually need them to be send over ws. Otherwise i guess it can be changed to logging pretty fast, right?

diamondburned:
> @diamondburned thats not exactly what i meant, sorry for misunderstanding.
> So commenting handling unknown events is definitely not what we need.
> I'm not sure whether we need those 2 events at all or they can be just log messages. First can be clarified with @iwasaki-kenta and @claudiucelfilip
> 
> But how do you think much time it will take to add ws support (parsing mainly) for those two events? Thats in case we actually need them to be send over ws. Otherwise i guess it can be changed to logging pretty fast, right?

Changing to logging is simple, it's just changing `log.Event()` to `log.Node()`. However, the current logging structure is ideally done over WS, so it should either be an actual event, or it should be a debugging event (which doesn't go over WS afaik).

That said, if these 2 events are needed over WS, I could do that.

a-urth:
@diamondburned so those events are both used in frontend, so we need to support them.

diamondburned:
> @diamondburned so those events are both used in frontend, so we need to support them.

It's added now.

log.go
```diff
@@ -129,6 +129,7 @@ func (c *CollapseResultsLogger) Log(results *collapseResults) {
 func (c *CollapseResultsLogger) addTx(mod, event []byte,
 	timestamp time.Time, tag int,
 	txID []byte, sender []byte, logError error) {
+
```
a-urth: remove this new line please

a-urth:
@diamondburned and merge master to this branch

Issue #416, opened 2019-12-11
Genesis test sync fail results to corrupt db files.


a-urth:
Due to how genesis test network setup is structured if there is timeout failure waiting for sync, `t.Fatal` will be raised and no db cleanup will be done.
https://github.com/perlin-network/wavelet/blob/master/genesis_integration_test.go#L83

a-urth:
I guess it makes sense to make overhaul to integration tests and remove all `t.Fatal` calls from helper function which can mess up tests execution flow.

PR #424, opened 2019-12-12
Refactor integration tests.

a-urth:
Started with #416, but ended up changing all integration tests.

Thats a questionable decision on my behalf but hear me out - my idea is not to change one style to another but rather clarify intention for tests. Previously we were passing `testing.T` to a lot of functions, ones were checking data and ones were `Fatal`ing in case of errors. Which was really hardening reading through tests and complicating writing new ones since it was not clear what will happen with `testing.T` in each method. And in each point of time you was not able to tell exactly will test execution flow will be affected or not. That particular problem caused bug in genesis test, which in case of `Fatal`, while waiting for consensus, led to cleanup function not being returned and not being called. In result database was corrupted and this is how i found it.

So I propose this way:
- Passing `testing.T` is totally fine but only if it validates i.e. checks data
- all error cases which needs to prevent tests from execution need to be handled in tests itself, only exception may be helper functions either really generic (like `FailTest` introduced in this pr) or specific for some set of tests

ADDITIONAL
I also simplified a lot concurrency used in methods which "wait" for some event.
And fixed some tests (reward and stake ones).
I removed insufficient funds tests, since we won't proceed to next round now with such transactions - they just won't be accepted by other nodes during tx sync.
I more or less unified all timeouts for wait methods.
Merkle root for tests is changed since I added rewards into genesis for test wallet2 in order to fix reward test.
Tests seems to be both more stable and faster now.

NOTE:
I'm going to change ALL
```golang
if !assert.NoError(b, err) {
	return
}
```
with
```go
FailTest(t, err)
```


hasyimibhar:
@a-urth I rebased this PR on top of master, because there is significant change on syncing due to introduction of `SyncManager`.

a-urth:
@hasyimibhar yes, syncing changed a lot, but it should've not affect tests that much. when i was checking it only `WaitForSync` should've change.

a-urth:
@hasyimibhar @iwasaki-kenta all test problems are resolved, this is may be reviewed and merged

PR #429, opened 2019-12-16
Add gossiping to share incoming transactions.

hasyimibhar:
This PR reintroduces gossiping using the same code from `v0.2.0`. The purpose is to compare its performance with the current transactions syncing using cuckoo filter.


rpc.proto
```diff
@@ -85,6 +92,7 @@ message TransactionPullResponse {
 }
 
 service Wavelet {
+		rpc Gossip (stream GossipRequest) returns (GossipResponse) {}
```
a-urth: do we need stream for that? how many transactions do we expect to send at once?

a-urth:
Syncing transactions is commented and all the related code left. There are 2 reasons why we still want to use it:
- to ensure all of the transactions are there, run it maybe once a minute
- to pull transactions for previous blocks on start - since now we only store (and get from peers as part of sync state) transaction ids for finalized blocks

